PR #6, opened 2022-11-18
feat: adding custom errors

destroyersrt:


LEAFERx (approved):
Fix several things due to open source migration. LGTM now. @ujvl can you double check?

contracts/src/libraries/Errors.sol
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+pragma solidity ^0.8.0;
+
```
ujvl, line 20: i think it might be more convenient to keep error definitions that are specific to certain functionality in the contract files they're meant for and use this file only for generic errors that are used across multiple contracts

LEAFERx: +1 @destroyersrt can you simply move these errors if only used by one component? thanks!

ujvl (approved):

PR #8, opened 2022-11-23
Added the archive flag to start an archive node

saxenism:
# Goals of PR

1. Added some clarity to the README files
2. Added the `gcmode archive` flag to the validator and sequencer startup scripts


LEAFERx:
We don't want our sequencer/validator to be archival by default. We should design another indexer service for the block explorer. Let's close it for now. For local testing you can copy-paste a new script e.g. `start_sequencer_archive.sh`.

PR #14, opened 2022-12-14
Inbox sync & indexer service

LEAFERx:
# Goals of PR

Dependencies:

- #11 
- #12 

Core changes:

- Refactor validator's inbox syncing, move to base service
- Validator behavior change: executes transactions when they are posted in the inbox instead of when new assertions published
- `BaseService.SyncInbox` allows a service to sync past sequenced transactions
- `BaseService.SyncLoop` allows a service to sync past sequenced transactions as well as watch ongoing ones
- `Indexer` is a service that only syncs sequenced transactions to its local blockchain, it won't participate in the validation/challenge
  - It is designed for the block explorer when used with archive flag

Notes:

- Assertion/Stake sync is not implemented in this pr so the sequencer/validator still can not be restarted

### DIFF
--- a/clients/geth/specular/rollup/services/indexer/indexer.go
+++ b/clients/geth/specular/rollup/services/indexer/indexer.go
@@ -1,0 +1,72 @@
+package indexer
+
+import (
+	bind "github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/log"
+	"github.com/ethereum/go-ethereum/node"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/specularl2/specular/clients/geth/specular/proof"
+	"github.com/specularl2/specular/clients/geth/specular/rollup/services"
+)
+
+func RegisterService(stack *node.Node, eth services.Backend, proofBackend proof.Backend, cfg *services.Config, auth *bind.TransactOpts) {
+	indexer, err := New(eth, proofBackend, cfg, auth)
+	if err != nil {
+		log.Crit("Failed to register the Rollup service", "err", err)
+	}
+	stack.RegisterLifecycle(indexer)
+	// stack.RegisterAPIs(indexer.APIs())
+	log.Info("Indexer registered")
+}
+
+type Indexer struct {
+	*services.BaseService
+
+	newBatchCh chan struct{}
+}
+
+func New(eth services.Backend, proofBackend proof.Backend, cfg *services.Config, auth *bind.TransactOpts) (*Indexer, error) {
+	base, err := services.NewBaseService(eth, proofBackend, cfg, auth)
+	if err != nil {
+		return nil, err
+	}
+	s := &Indexer{
+		BaseService: base,
+		newBatchCh:  make(chan struct{}, 1),
+	}
+	return s, nil
+}
+
+func (i *Indexer) newBatchConsumeLoop() {
+	defer i.Wg.Done()
+	for {
+		select {
+		case <-i.newBatchCh:
+			continue
+		case <-i.Ctx.Done():
+			return
+		}
+	}
+}
+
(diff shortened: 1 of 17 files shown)


silmaj (approved):

clients/geth/specular/rollup/services/indexer/indexer.go
```diff
@@ -0,0 +1,72 @@
...
+	indexer, err := New(eth, proofBackend, cfg, auth)
+	if err != nil {
+		log.Crit("Failed to register the Rollup service", "err", err)
+	}
+	stack.RegisterLifecycle(indexer)
+	// stack.RegisterAPIs(indexer.APIs())
+	log.Info("Indexer registered")
+}
+
+type Indexer struct {
```
ujvl, line 22: can we just fold this functionality into `validator.go` instead of having a separate class? just add some feature flagging to turn off assertion creation / challenging (just dont start the loop i guess); and then in service.go just switch on `cfg.Node` with a common case for `services.{NODE_VALIDATOR, NODE_INDEXER}`

i think that would improve maintainability / get rid of some common code

LEAFERx: we can list it as a separate task

clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -204,31 +119,58 @@ func (v *Validator) validationLoop(genesisRoot common.Hash) {
...
 		PrevCumulativeGasUsed: new(big.Int),
 	}
+	// The next assertion to be validated
+	var currentAssertion *rollupTypes.Assertion
 
 	isInChallenge := false
 
+	validateCurrentAssertion := func() error {
+		// Validate current assertion
+		err := v.tryVaidateAssertion(confirmedAssertion, currentAssertion)
```
ujvl, line 129:
```suggestion
		err := v.tryValidateAssertion(confirmedAssertion, currentAssertion)
```

clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -204,31 +119,58 @@ func (v *Validator) validationLoop(genesisRoot common.Hash) {
 		Deadline:              new(big.Int),
 		PrevCumulativeGasUsed: new(big.Int),
 	}
+	// The next assertion to be validated
+	var currentAssertion *rollupTypes.Assertion
 
 	isInChallenge := false
 
+	validateCurrentAssertion := func() error {
```
ujvl, line 127: why not move this out as a separate method instead of using a lambda

clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -42,141 +42,58 @@ func New(eth services.Backend, proofBackend proof.Backend, cfg *services.Config,
...
-		var receipts []*types.Receipt
-		for idx, tx := range sblock.Txs {
-			state.Prepare(tx.Hash(), idx)
-			receipt, err := core.ApplyTransaction(chainConfig, v.Chain, &v.Config.SequencerAddr, gasPool, state, header, tx, &header.GasUsed, *v.Chain.GetVMConfig())
-			if err != nil {
-				return common.Hash{}, nil, err
-			}
-			receipts = append(receipts, receipt)
+// This function will try to validate a pending assertion
+func (v *Validator) tryVaidateAssertion(confirmedAssertion, assertion *rollupTypes.Assertion) error {
```
ujvl, line 53:
```suggestion
func (v *Validator) tryValidateAssertion(confirmedAssertion, assertion *rollupTypes.Assertion) error {
```
also nit maybe we should call this `locallyConfirmedAssertion` (to disambiguate from confirmation in the L1 sense which implies finality)

ujvl (approved):

PR #16, opened 2022-12-21
use blocks instead of batches in sequencer

kwhuo68:
Core changes:
-Use blocks instead of batches in sequencer

### DIFF
--- a/clients/geth/specular/rollup/services/sequencer/sequencer.go
+++ b/clients/geth/specular/rollup/services/sequencer/sequencer.go
@@ -187,18 +187,6 @@ func (s *Sequencer) batchingLoop() {
 	}
 }
 
-// Combines batches
-func combineBatches(slice []*rollupTypes.TxBatch) *rollupTypes.TxBatch {
-	var blocks []*types.Block
-
-	for i := 0; i <= len(slice)-1; i++ {
-		currBlocks := slice[i].Blocks
-		blocks = append(blocks, currBlocks...)
-	}
-	combinedBatch := rollupTypes.NewTxBatch(blocks, 0)
-	return combinedBatch
-}
-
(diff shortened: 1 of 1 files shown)


clients/geth/specular/rollup/services/sequencer/sequencer.go
```diff
@@ -53,7 +53,7 @@ func New(eth services.Backend, proofBackend proof.Backend, cfg *services.Config,
 	}
 	s := &Sequencer{
 		BaseService:          base,
-		batchCh:              make(chan *rollupTypes.TxBatch, 4096),
+		blockCh:              make(chan *types.Blocks, 4096),
```
LEAFERx, line 56: `types.Blocks` is a slice, i.e. ref type. you shouldn't use a pointer here

LEAFERx (changes requested):

LEAFERx (approved):

clients/geth/specular/rollup/services/sequencer/sequencer.go
```diff
@@ -277,8 +276,9 @@ func (s *Sequencer) sequencingLoop(genesisRoot common.Hash) {
 				commitAssertion()
 			}
 			batchTxs = nil
-		case batch := <-s.batchCh:
+		case blocks := <-s.blockCh:
 			// New batch from Batcher
+			batch := rollupTypes.NewTxBatch(*blocks, 0) // TODO: add max batch size
```
LEAFERx, line 281: shouldn't create batches here; instead please push these blocks into a local array. create batches in ticks

LEAFERx (changes requested):

ujvl (approved):

PR #18, opened 2022-12-28
[feature] Add staking event

ujvl:
# Goals of PR

Provides stakers an on-chain checkpoint `event StakerStaked` when they advance their stake or create a new assertion (useful since idle stakers can be removed from the contract state w/o their permission)

Only 2 LoC changed in contracts here ^ -- rest of diff is misc:
- Update bindings, seems like they were out-of-date; maybe we should add as a step in precommit to make sure they stay up-to-date (or remove from version control altogether?)
- Consolidate scripts under `./sbin/` (todo: move rest of testing scripts here too)
- Add `install` target to Makefile (generates bindings and installs geth)


clients/geth/specular/Makefile
```diff
@@ -1,8 +1,13 @@
...
+.PHONY: geth bindings install
 
 GOBIN = ./build/bin
 
+install: bindings geth
+
+bindings:
+	go generate ./...
+
 geth:
```
LEAFERx, line 10: geth should depend on bindings

LEAFERx (approved):

PR #19, opened 2022-12-29
[feature] Sync validator/sequencer state from L1

ujvl:
# Goals of PR

- Syncs last validated DA, so a restarted validator doesn't need to re-validate DAs. 
- Syncs previously sequenced txs on new run
- Adds index to two events (StakerStaked and AssertionCreated) for query convenience.
- Cleanup + refactor functionality for ease of unit testing (thread-safe `L1BridgeClient` abstracts out L1 communication)
- `context.Context` plumbing -- top-level initialization (in BaseService), plumbed down through methods instead of state

TODO:
- fix stake advancement when local last validated and last staked are out of sync


silmaj (approved):

LEAFERx (changes requested):
sequencer's assertion state also needs to be synced in `sequencingLoop`

LEAFERx (approved):

PR #20, opened 2022-12-30
[fix] fix validation retry

LEAFERx:
# Goals of PR

Core changes:

- Fix behavior when validation failed
  - Challenge when vmhash or gas mismatch
  - Retry when inbox not fully synced
  - Panic when other errors happen
  - TODO: recover from other errors and retry

### DIFF
--- a/clients/geth/specular/rollup/services/validator/validator.go
+++ b/clients/geth/specular/rollup/services/validator/validator.go
@@ -135,7 +135,8 @@ func (v *Validator) validationLoop(genesisRoot common.Hash) {
 			case errors.Is(err, errAssertionOverflowedInbox):
 				// Assertion overflowed inbox, wait for next block
 			}
-			return err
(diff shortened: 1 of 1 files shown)


clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -132,12 +132,15 @@ func (v *Validator) validationLoop(genesisRoot common.Hash) {
 			case errors.Is(err, errValidationFailed):
 				// Validation failed, challenge
 				isInChallenge = true
+				return nil
 			case errors.Is(err, errAssertionOverflowedInbox):
```
ujvl, line 136: i'd log something here at least (or before returning the err above). also consider changing the name (e.g. `errPendingSync`) since it's a bit unclear whether it's overflowing the L1 inbox or L2 blockchain current state

ujvl (approved):

PR #22, opened 2023-01-05
make data dir and move scripts

kwhuo68:
# Goals of PR

Core changes:
-Make main data dir
-Get rid of relative paths 
-Move test-node/ scripts to sbin/ and modify accordingly


clients/geth/specular/sbin/import_accounts.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/bash
+SBIN=`dirname $0`
+SBIN="`cd "$SBIN"; pwd`"
+. $SBIN/configure.sh
+cd $DATA_DIR
```
ujvl, line 5: are all these scripts supposed to be run w/ `source` (`. my_script.sh`) or `./my_script.sh`? if the former, i’d avoid `cd` here and just use the absolute paths below (i.e. `$GETH_SPECULAR_DIR/build/bin/geth` etc)

same question for the other scripts

clients/geth/specular/sbin/import_accounts.sh
```diff
@@ -0,0 +1,11 @@
...
+../build/bin/geth --password ./password.txt --datadir ./data account import ./keys/sequencer.prv
```
ujvl, line 6: can we make this `data_sequencer`; otherwise we got the ugly `data/data`

ujvl (commented):
looks good just some minor comments!

ujvl (approved):

PR #23, opened 2023-01-06
Changing Proxy Pattern to UUPS

destroyersrt:


LEAFERx (changes requested):
You also need to modify the contract to be `UUPSUpgradeable`.

LEAFERx (approved):

PR #37, opened 2023-02-02
[WIP] Rollup test file

saxenism:


destroyersrt:
[AssertionOutOfRange]( https://github.com/SpecularL2/specular/blob/rahul/testing-alpha/contracts/src/Rollup.sol#L197 ) error naming 
```solidity
if (assertionID <= staker.assertionID || assertionID > lastCreatedAssertionID) {
            revert AssertionOutOfRange();
```
I think it is alright to keep the error AssertionOutOfRange  as it doesn't lie in the valid assertion Id range.

destroyersrt:
[challengeAssertion](https://github.com/SpecularL2/specular/blob/rahul/testing-alpha/contracts/src/Rollup.sol#L313) - defender & Challenger being same address
```solidity
function challengeAssertion(address[2] calldata players, uint256[2] calldata assertionIDs)
        external
        override
        returns (address)
    {
        uint256 defenderAssertionID = assertionIDs[0];
        uint256 challengerAssertionID = assertionIDs[1];
        // Require IDs ordered and in-range.
        if (defenderAssertionID >= challengerAssertionID) {
            revert WrongOrder();
        }
        if (challengerAssertionID > lastCreatedAssertionID) {
            revert UnproposedAssertion();
        }
        if (lastConfirmedAssertionID >= defenderAssertionID) {
            revert AssertionAlreadyResolved();
        }
...
```
Challenger and Defender being the same address in challengeAssertion , one of the two conditions will eventually fail, so we can keep it the way it is right now, and anyway it won't be feasible for the user to do this sort of thing.

saxenism:
This PR is deprecated in favour of individual modular PRs adding testing for specific functions of `Rollup.sol`

PR #53, opened 2023-02-21
Osp

silmaj:
# Goals of PR

Core changes:

- Describe changes here

Notes:

- Write notes here


clients/geth/specular/rollup/services/sequencer/sequencer.go
```diff
@@ -335,7 +333,7 @@ func (s *Sequencer) confirmationLoop() {
 
 	// Current pending assertion from sequencing goroutine
 	// TODO: watch multiple pending assertions
-	var pendingAssertion *rollupTypes.Assertion
+	pendingAssertion := new(rollupTypes.Assertion)
```
ujvl, line 336: this is equivalent right

ujvl (commented):
we'll merge `main` into `osp` which addresses 2/3 of these issues. we can just limit this PR to removing l2 gas (entirely)—also this requires some changes on L2 to work

ujvl:
closing since we get rid of l2 gas entirely in PR 89

PR #55, opened 2023-02-23
storage-layout generation

destroyersrt:
This PR adds: 
- script to generate the storage layout
- storage layout of current contracts

### DIFF
--- a/contracts/.storage-layout
+++ b/contracts/.storage-layout
@@ -1,0 +1,92 @@
+=======================
+👁👁 STORAGE LAYOUT snapshot 👁👁
+=======================
+
+=======================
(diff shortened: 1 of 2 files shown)


contracts/.storage-layout
```diff
@@ -0,0 +1,92 @@
+=======================
```
LEAFERx, line 1: either add this file to gitigore or add precommit/ci pipeline to auto generate this

LEAFERx (changes requested):
this file and script is not helpful. we don't want to manually assign storages by checking this file, instead we want a structured mapping that deployment scrip can access to initialize storage using variable names directle

LEAFERx:
Deprecated by #117

PR #62, opened 2023-02-27
Add docker and blockscout

LEAFERx:
# Goals of PR

Core changes:

- Add L1 connection retry in `BaseService::NewBaseService` to allow the node to wait on the l1 start in integration test environment
- Dockerize l2 geth
- Dockerize l1 hardhat
- Add blockscout subdmoule
- Add docker compose files under `docker` for different services


clients/geth/specular/Makefile
```diff
@@ -4,10 +4,18 @@ GOBIN = ./build/bin
 
 install: bindings geth
 
-bindings:
+compile-contracts:
```
ujvl, line 7: minor nit: just `contracts` instead of `compile-contracts` (compile implied by make action)

clients/geth/specular/data/.gitignore
```diff
@@ -1,4 +1,5 @@
 data
 data_sequencer
 data_validator
-data_indexer
\ No newline at end of file
+data_indexer
```
ujvl, line 4: think we can get rid of all the `data_*` from here since they're under `data/` now

clients/geth/specular/Dockerfile
```diff
@@ -0,0 +1,30 @@
+FROM node:16-alpine3.17 as contracts-builder
```
ujvl, line 1: nit: some comments in this file might be helpful

clients/geth/specular/rollup/services/base.go
```diff
@@ -33,12 +35,31 @@ type BaseService struct {
 	Wg     sync.WaitGroup
 }
 
+func connectL1(ctx context.Context, endpoint string) (*ethclient.Client, error) {
+	var l1 *ethclient.Client
+	var err error
+	retryOpts := []retry.Option{
+		retry.Context(ctx),
+		retry.Attempts(3),
```
ujvl, line 43: we could also consider just letting it retry perpetually every x seconds until it connects

docker/scripts/geth.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
+
+geth --datadir . --password ./password.txt account import ./key.prv
+
+geth --datadir . --networkid $NETWORK_ID init ./genesis.json
+
+exec geth \
```
ujvl, line 7: why `exec` here

contracts/scripts/create_genesis.ts
```diff
@@ -70,7 +70,8 @@ const main = () => {
 
   if (inFlagIndex > -1) {
     baseGenesisPath = process.argv[inFlagIndex + 1];
-    GenesisJson = require(baseGenesisPath);
+    GenesisJson = JSON.parse(fs.readFileSync(baseGenesisPath, "utf-8"));
+    // GenesisJson = require(baseGenesisPath);
```
ujvl, line 74: nit: avoid commented lines if possible (just delete)

clients/geth/specular/rollup/services/base.go
```diff
@@ -33,12 +35,31 @@ type BaseService struct {
...
+		retry.LastErrorOnly(true),
+		retry.OnRetry(func(n uint, err error) {
+			log.Error("Failed to connect to L1", "endpoint", endpoint, "attempt", n, "err", err)
+		}),
+	}
+	err = retry.Do(func() error {
+		l1, err = ethclient.DialContext(ctx, endpoint)
+		return err
+	}, retryOpts...)
+	return l1, err
```
ujvl, line 54: nit: think the convention we should try to maintain is to wrap all errors w/ additional context before returning
```suggestion
if err != nil {
    return nil, fmt.Errorf("Could not connect to L1, err: %w", err)
}
return l1, nil
```

clients/geth/specular/sbin/start_indexer.sh
```diff
@@ -17,6 +18,7 @@ $GETH_SPECULAR_DIR/build/bin/geth \
     --rollup.coinbase=f39fd6e51aad88f6f4ce6ab8827279cfffb92266 \
     --rollup.l1endpoint 'ws://localhost:8545' \
     --rollup.l1chainid 31337 \
+    --rollup.sequencer-addr '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266' \
     --rollup.sequencer-inbox-addr '0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512' \
     --rollup.rollup-addr '0x5FC8d32690cc91D4c39d9d3abcBD16989F875707' \
```
ujvl, line 23: maybe orthogonal to your PR but do these need to be hardcoded here—we're dumping them to a file now right? can we read from that

LEAFERx: yes we can make an env file for that and actually we should do that for both this and docker compose file, but maybe in next pr

docker/scripts/geth.sh
LEAFERx: this will swap `sh` command to `geth`

ujvl (approved):

clients/geth/specular/rollup/services/base.go
LEAFERx: its not friendly to ci/cd if l1 fails

ujvl: yeah actually what you have makes more sense

ayushguptacse14 (approved):

PR #63, opened 2023-02-27
Fold AssertionMap into Rollup.sol and other misc changes

ujvl:
# Goals of PR

Core changes:

- Remove AssertionMap
- Remove stakeToken param
- Add params for initializing rollup
- Expose Staker struct and Assertion (separated mappings out into `AssertionState`), and return that instead of individual params
- Some error renaming


LEAFERx (approved):

LEAFERx:
lgtm

PR #83, opened 2023-03-06
[Testing] Rollup Tests: unstake

saxenism:
# Goals of PR

Tests for function `unstake` in Rollup.sol


contracts/test/Rollup.t.sol
```diff
@@ -0,0 +1,426 @@
...
+        alice = users[1];
+        vm.label(alice, "Alice");
+
+        bob = users[2];
+        vm.label(bob, "Bob");
+
+        vault = makeAddr("vault");
+        sequencerOwner = makeAddr("sequencerOwner");
+        sequencerAddress = makeAddr("sequencerAddress");
+        rollupOwner = makeAddr("rollupOwner");
```
LEAFERx, line 65: why two ways of creating users? please clean up

LEAFERx (changes requested):

contracts/test/Rollup.t.sol
saxenism: The reason was it was easier for me to move forward by just using the cheat `makeAddr`,  since that would mean no extra *utils* file required.
But for some reason, it was not possible for me to change their Ether balance. Therefore, I have reverted back to the old method of creating new users.

Also removed the SpecularProxy.

Please review again. Thank you.

PR #85, opened 2023-03-07
Add Assertion Rejection to sequencer

ayushguptacse14:
# Goals of PR

Core changes:

- Add Assertion Rejection to the sequencer

Notes:

- RejectFirstUnresolvedAssertion(s.Config.SequencerAddr) is called for Sequencer's Address as the sequencer creates all assertions and on creating an assertion it also stakes on assertion. (In Rollup.sol: createAssertion -> stakeOnAssertion -> StakerStaked)

### DIFF
--- a/clients/geth/specular/rollup/services/sequencer/sequencer.go
+++ b/clients/geth/specular/rollup/services/sequencer/sequencer.go
@@ -339,14 +339,6 @@ func (s *Sequencer) sequencingLoop(ctx context.Context) {
 func (s *Sequencer) confirmationLoop(ctx context.Context) {
 	defer s.Wg.Done()
 
-	// Watch AssertionRejected event
-	rejectedCh := make(chan *bindings.IRollupAssertionRejected, 4096)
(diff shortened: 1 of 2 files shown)


clients/geth/specular/rollup/services/sequencer/sequencer.go
```diff
@@ -399,6 +407,22 @@ func (s *Sequencer) confirmationLoop(ctx context.Context) {
...
 					}
 				}
+			case <-rejectedCh:
+				if !pendingConfirmed {
+					// Reject the first unresolved assertion if all the conditions are met
+					// (1) challenge period has passed, 
+					// (2) at least one staker exists, 
+					// (3) no staker remains staked on the assertion (all have been destroyed).
+					
+					_, err := s.L1Client.RejectFirstUnresolvedAssertion(s.Config.SequencerAddr)
```
ujvl, line 417: this doesn't make sense; you're calling reject when you get a reject event?

ayushguptacse14: Modifications have been implemented according to feedback, however, rejection criteria is yet to be determined.

PR #39, opened 2023-02-09
Custom Error and Custom Logs Added

ayushguptacse14:
# Goals of PR

Core changes:

- Error returned upstream
- Fetching function name from runtime 

Error Logs generated are as follows:
```
CRIT [03-02|17:13:36.576] Failed to register the Rollup service%!(EXTRA string=err, *fmt.Error=*errors.errorString NewBaseService Failed to Start%!(EXTRA rpc.wsHandshakeError=dial tcp [::1]:8545: connect: connection refused)
/Users/ayush/Documents/GitHub/specular/clients/geth/specular/rollup/utils/fmt/fmt.go:21 (0x1017b9f30)
        Errorf: err:  errors.Errorf(format, args...),
 | /Users/ayush/Documents/GitHub/specular/clients/geth/specular/rollup/services/sequencer/sequencer.go:54 (github.com/specularl2/specular/clients/geth/specular/rollup/services/sequencer.New)) | /Users/ayush/Documents/GitHub/specular/clients/geth/specular/rollup/services/sequencer/sequencer.go:28 (github.com/specularl2/specular/clients/geth/specular/rollup/services/sequencer.RegisterService) 
ayush@Ayushs-MacBook-Air-2 sbin % 
```

PR #86, opened 2023-03-07
Custom Error Added

ayushguptacse14:
# Goals of PR

Core changes:

- Old [PR 39](https://github.com/SpecularL2/specular/pull/39)
- Error returned upstream
- Fetching function name from runtime 

For a function with main calling a, a calling b & b calling c [main()->a->b->c] Error Logs generated are as follows:

```
main:11 | in main%!(EXTRA string=err, *fmt.wrapError=*fmt.wrapError assertion failed: *fmt.wrapError assertion failed: *errors.errorString assertion failed
/Users/ayush/hello/utils/fmt/fmt.go:18 (0x1007a132c)
        Errorf: err := errors.Errorf(format, args...)
 | main.c:32
/Users/ayush/hello/utils/fmt/fmt.go:18 (0x1007a132c)
        Errorf: err := errors.Errorf(format, args...)
 | main.b:26
/Users/ayush/hello/utils/fmt/fmt.go:18 (0x1007a132c)
        Errorf: err := errors.Errorf(format, args...)
 | main.a:18)0[0/0]0x02exit status 1
```

### DIFF
--- a/clients/geth/specular/rollup/utils/fmt/fmt.go
+++ b/clients/geth/specular/rollup/utils/fmt/fmt.go
@@ -3,28 +3,26 @@ package fmt
 import (
 	"fmt"
 	"runtime"
-
+	
 	"github.com/specularl2/specular/clients/geth/specular/rollup/utils/errors"
 )
 
-type Error struct {
+type wrapError struct {
 	err  *errors.Error
-	file string
-	line int
-	fn   *runtime.Func
+	msg string
 }
 
-func Errorf(format string, args ...interface{}) *Error {
-	pc, file, line, _ := runtime.Caller(1)
+func Errorf(format string, args ...interface{}) *wrapError {
+	pc, _, line, _ := runtime.Caller(1)
 	fn := runtime.FuncForPC(pc)
-	return &Error{
-		err:  errors.Errorf(format, args...),
-		file: file,
-		line: line,
-		fn:   fn,
+	err := errors.Errorf(format, args...)
+	return &wrapError{
+		err:  err,
+		msg: fmt.Sprintf("%s | %s:%d", err.ErrorStack(), fn.Name(), line),
 	}
 }
 
-func (e *Error) Error() string {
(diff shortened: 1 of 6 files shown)


clients/geth/specular/rollup/utils/fmt/fmt.go
```diff
@@ -0,0 +1,28 @@
...
+type wrapError struct {
+	err  *errors.Error
+	msg string
+}
+
+func Errorf(format string, args ...interface{}) *wrapError {
+	pc, _, line, _ := runtime.Caller(1)
+	fn := runtime.FuncForPC(pc)
+	err := errors.Errorf(format, args...)
+	return &wrapError{
```
ujvl, line 19: do we really need to wrap the error here—`errors.Errorf` already wraps it; can't we just concat with `format` before passing it in L18?

ujvl: actually on second thought, looking at the output the information it's appending here seems completely redundant; e.g.
```
/Users/ayush/hello/main.go:35 (0x100a8fadc)
        c: return fmt.Errorf("assertion failed for c: %w", err)
 | main.c:35                      <---- this is already mentioned on the first line
```

i suggest getting rid of everything here and renaming the `errors` package to `fmt` so we can just call `Errorf` directly from there

ujvl: ah got it can you just modify directly:
```
func (err *Error) Error() string {

	msg := err.Err.Error()
	if err.prefix != "" {
		msg = fmt.Sprintf("%s: %s", err.prefix, msg)
	}
	<specular modification>
	msg = msg + "\n" + string(err.Stack())
	<specular modification/>
	return msg 
}
```
and maybe remove ErrorStack then since it's redundant

and in the files borrowed we should attribute to go-errors (can you find what the correct licensing text is, their repo is MIT i believe)

ayushguptacse14: The changes have been implemented successfully.

ujvl (approved):

PR #88, opened 2023-03-15
Make Sequencer and Validator start with current L1RollupBlock

ayushguptacse14:
# Goals of PR

Core changes:

- Made Sequencer and Validator start with L1RollupBlock
- On starting the L2 sequencer when we sync to the L1 head, our processTxBatchAppendedEvents will create blocks from genesis but commit only the blocks ahead of the current L2 block.


clients/geth/specular/rollup/services/base.go
```diff
@@ -65,6 +61,26 @@ func (b *BaseService) Chain() *core.BlockChain {
 	return b.Eth.BlockChain()
 }
 
+
+func (b *BaseService) GetL1RollupBlock(ctx context.Context, start uint64) uint64 {
+	l1blocknumber := start
+	l1BlockHead, err := b.L1Client.BlockNumber(ctx)
+	if err != nil {
+		return b.Config.L1RollupGenesisBlock
```
ujvl, line 69: why return this instead of an error?

clients/geth/specular/rollup/services/base.go
```diff
@@ -65,6 +61,26 @@ func (b *BaseService) Chain() *core.BlockChain {
...
+	}
+	opts := bind.FilterOpts{Start: b.Config.L1RollupGenesisBlock, End: &l1BlockHead, Context: ctx}
+	eventsIter, err := b.L1Client.FilterTxBatchAppendedEvents(&opts)
+	if err != nil {
+		return b.Config.L1RollupGenesisBlock
+	}
+	for eventsIter.Next() {
+		if eventsIter.Event.Raw.BlockNumber > l1blocknumber{
```
ujvl, line 77: i don't really understand what the intent is here, you're just getting the latest tx batch's starting block number; that doesn't tell you anything about whether it corresponds to the local state

also please apply go format (can you enable this in your IDE? its automatic in vscode)

clients/geth/specular/rollup/services/base.go
```diff
@@ -65,6 +61,26 @@ func (b *BaseService) Chain() *core.BlockChain {
 	return b.Eth.BlockChain()
 }
 
+
+func (b *BaseService) GetL1RollupBlock(ctx context.Context, start uint64) uint64 {
```
ujvl, line 65: this should be something like
```suggestion
func (b *BaseService) GetLatestL1BlockForL2Block(ctx context.Context, l2Block uint64, start uint64) uint64 {
```

ujvl (approved):

ujvl (approved):

PR #92, opened 2023-03-22
Setters and tests for parameters in RollupBase

Andy-Zhou2:
# Added setters and tests for parameters in RollupBase

Core changes:

- new event `ConfigurationChanged()`
- setter functions
- tests

Notes:

- `baseStakeAmount `can only be decreased

### DIFF
--- a/contracts/src/IRollup.sol
+++ b/contracts/src/IRollup.sol
@@ -33,6 +33,8 @@ interface IRollup {
 
     event StakerStaked(address stakerAddr, uint256 assertionID);
 
+    event ConfigurationChanged();
+
     // TODO: Include errors thrown in function documentation.
 
     /// @dev Thrown when assertion creation requested with invalid inbox size.
@@ -114,6 +116,9 @@ interface IRollup {
 
     /// @dev Thrown when there are zero stakers
     error NoStaker();
+
+    /// @dev Thrown when the staker amount is set to be increased
+    error IncreaseStake();
 
     struct Staker {
         bool isStaked;
@@ -237,4 +242,32 @@ interface IRollup {
      * assertion is not the last confirmed assertion.
      */
     function rejectFirstUnresolvedAssertion(address stakerAddress) external;
+
+    /**
+     * @notice Sets the confirmation period
+     * @param newPeriod New confirmation period
+     *
+     */
+    function setConfirmationPeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the challenge period
+     * @param newPeriod New challenge period
+     *
+     */
+    function setChallengePeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the minimum assertion period
+     * @param newPeriod New minimum assertion period
+     *
+     */
+    function setMinimumAssertionPeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the base stake amount
+     * @param newAmount New base stake amount, this can only be decreased
+     *
+     */
+    function setBaseStakeAmount(uint256 newAmount) external;
 }
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -50,8 +50,6 @@ abstract contract RollupBase is
     IDAProvider public daProvider;
     IVerifier public verifier;
 
-    event ConfigurationChanged();
-
     struct AssertionState {
         mapping(address => bool) stakers; // all stakers that have ever staked on this assertion.
         mapping(bytes32 => bool) childStateHashes; // child assertion vm hashes
@@ -67,24 +65,28 @@ abstract contract RollupBase is
         __UUPSUpgradeable_init();
     }
 
+    /// @inheritdoc IRollup
     function setConfirmationPeriod(uint256 newPeriod) public onlyOwner {
         confirmationPeriod = newPeriod;
         emit ConfigurationChanged();
     }
 
+    /// @inheritdoc IRollup
     function setChallengePeriod(uint256 newPeriod) public onlyOwner {
         challengePeriod = newPeriod;
         emit ConfigurationChanged();
     }
 
+    /// @inheritdoc IRollup
     function setMinimumAssertionPeriod(uint256 newPeriod) public onlyOwner {
         minimumAssertionPeriod = newPeriod;
         emit ConfigurationChanged();
     }
 
+    /// @inheritdoc IRollup
     function setBaseStakeAmount(uint256 newAmount) public onlyOwner {
         if (newAmount > baseStakeAmount) {
-            revert("Cannot increase base stake amount");
+            revert IncreaseStake();
         }
 
         baseStakeAmount = newAmount;
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -71,6 +71,25 @@ abstract contract RollupBase is
         confirmationPeriod = newPeriod;
         emit ConfigurationChanged();
     }
+
+    function setChallengePeriod(uint256 newPeriod) public onlyOwner {
+        challengePeriod = newPeriod;
+        emit ConfigurationChanged();
+    }
+
+    function setMinimumAssertionPeriod(uint256 newPeriod) public onlyOwner {
+        minimumAssertionPeriod = newPeriod;
+        emit ConfigurationChanged();
+    }
+
+    function setBaseStakeAmount(uint256 newAmount) public onlyOwner {
+        if (newAmount > baseStakeAmount) {
+            revert("Cannot increase base stake amount");
+        }
+
+        baseStakeAmount = newAmount;
+        emit ConfigurationChanged();
+    }
 }
 
 contract Rollup is RollupBase {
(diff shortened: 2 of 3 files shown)


contracts/src/Rollup.sol
```diff
@@ -64,6 +66,30 @@ abstract contract RollupBase is
...
+    }
+
+    function setMinimumAssertionPeriod(uint256 newPeriod) public onlyOwner {
+        minimumAssertionPeriod = newPeriod;
+        emit ConfigurationChanged();
+    }
+
+    function setBaseStakeAmount(uint256 newAmount) public onlyOwner {
+        if (newAmount > baseStakeAmount) {
+            revert("Cannot increase base stake amount");
```
ujvl, line 87: use an `error` definition instead of string (see `IRollup.sol`)

contracts/src/IRollup.sol
```diff
@@ -237,4 +242,32 @@ interface IRollup {
      * assertion is not the last confirmed assertion.
      */
     function rejectFirstUnresolvedAssertion(address stakerAddress) external;
+
+    /**
+     * @notice Sets the confirmation period
+     * @param newPeriod New confirmation period
+     *
```
saxenism, line 249: Remove this extra line from natspec

contracts/src/IRollup.sol
```diff
@@ -237,4 +242,32 @@ interface IRollup {
...
+     */
+    function setConfirmationPeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the challenge period
+     * @param newPeriod New challenge period
+     *
```
saxenism, line 256: Remove extra line from natspec.

contracts/src/IRollup.sol
```diff
@@ -237,4 +242,32 @@ interface IRollup {
...
+     */
+    function setChallengePeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the minimum assertion period
+     * @param newPeriod New minimum assertion period
+     *
```
saxenism, line 263: Remove extra line from natspec

contracts/src/IRollup.sol
```diff
@@ -237,4 +242,32 @@ interface IRollup {
...
+     */
+    function setMinimumAssertionPeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the base stake amount
+     * @param newAmount New base stake amount, this can only be decreased
+     *
```
saxenism, line 270: Remove extra line from natspec.

contracts/src/IRollup.sol
```diff
@@ -237,4 +242,32 @@ interface IRollup {
...
+    /**
+     * @notice Sets the minimum assertion period
+     * @param newPeriod New minimum assertion period
+     *
+     */
+    function setMinimumAssertionPeriod(uint256 newPeriod) external;
+
+    /**
+     * @notice Sets the base stake amount
+     * @param newAmount New base stake amount, this can only be decreased
```
saxenism, line 269: @param Should only include what the parameter means.
Shift the *this can only be decreased* line to a separate line under @dev

contracts/src/IRollup.sol
```diff
@@ -115,6 +117,9 @@ interface IRollup {
     /// @dev Thrown when there are zero stakers
     error NoStaker();
 
+    /// @dev Thrown when the staker amount is set to be increased
+    error IncreaseStake();
```
saxenism, line 121: This name sounds misleading to me. Rename it to something that makes it even more easier to understand, for example: **StakeCannotBeIncreased**.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
         __Ownable_init();
         __UUPSUpgradeable_init();
     }
+
+    /// @inheritdoc IRollup
+    function setConfirmationPeriod(uint256 newPeriod) public override onlyOwner {
+        confirmationPeriod = newPeriod;
```
saxenism, line 70: Please introduce sanity checks here.
The `newPeriod` should be in a range between `minConfirmationPeriod` and `maxConfirmationPeriod`.

If nothing, make sure that `newPeriod` != 0.

However, I think we will first need to discuss what kind of checks do we need to implement if at all we need to implement them.

cc: @ujvl @LEAFERx

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+        emit ConfigurationChanged();
+    }
+
+    /// @inheritdoc IRollup
+    function setChallengePeriod(uint256 newPeriod) public override onlyOwner {
+        challengePeriod = newPeriod;
```
saxenism, line 76: No sanity checks.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+        emit ConfigurationChanged();
+    }
+
+    /// @inheritdoc IRollup
+    function setMinimumAssertionPeriod(uint256 newPeriod) public override onlyOwner {
+        minimumAssertionPeriod = newPeriod;
```
saxenism, line 82: No Sanity checks,

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+        emit ConfigurationChanged();
+    }
+
+    /// @inheritdoc IRollup
+    function setBaseStakeAmount(uint256 newAmount) public override onlyOwner {
+        if (newAmount > baseStakeAmount) {
```
saxenism, line 88: This check basically allows the owner to set the newAmount exactly what it was earlier. In which case, emitting the event `IncreaseStake` is counter-intuitive.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+            revert IncreaseStake();
+        }
+
+        baseStakeAmount = newAmount;
+        emit ConfigurationChanged();
```
saxenism, line 93: This event does not help anyone much. Consider adding the `previousValue` and `changedValue` in the event.

contracts/test/RollUp.t.sol
```diff
@@ -0,0 +1,236 @@
...
+            )
+        );
+        vm.stopPrank();
+
+        // Making sure that the proxy returns the correct proxy owner and sequencerAddress
+        address sequencerInboxDeployer = seqIn.owner();
+        assertEq(sequencerInboxDeployer, deployer);
+
+        address fetchedSequencerAddress = seqIn.sequencerAddress();
+        assertEq(fetchedSequencerAddress, sequencerAddress);
```
saxenism, line 87: Can remove these assertions. However, it is not necessary.

saxenism:
This is great work @Andy-Zhou2 , however I would like to understand the rationale behind introducing all these setters when `Rollup.sol` has the `init` function to set them.
Secondly, I am not very sure if the errors and events in `RollupBase.sol` should be included in `IRollup.sol` or in a separate `IRollupBase.sol`. Looping in @ujvl for comments on this.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
         __Ownable_init();
         __UUPSUpgradeable_init();
     }
+
+    /// @inheritdoc IRollup
+    function setConfirmationPeriod(uint256 newPeriod) public override onlyOwner {
```
saxenism, line 69: Change the function from public to external perhaps? That is the function signature in interface as well as the gas-efficient way to move forward assuming this function is not called inside the contracts.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+        confirmationPeriod = newPeriod;
+        emit ConfigurationChanged();
+    }
+
+    /// @inheritdoc IRollup
+    function setChallengePeriod(uint256 newPeriod) public override onlyOwner {
```
saxenism, line 75: Change the function from public to external perhaps? That is the function signature in interface as well as the gas-efficient way to move forward assuming this function is not called inside the contracts.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+        challengePeriod = newPeriod;
+        emit ConfigurationChanged();
+    }
+
+    /// @inheritdoc IRollup
+    function setMinimumAssertionPeriod(uint256 newPeriod) public override onlyOwner {
```
saxenism, line 81: Change the function from public to external perhaps? That is the function signature in interface as well as the gas-efficient way to move forward assuming this function is not called inside the contracts.

contracts/src/Rollup.sol
```diff
@@ -64,6 +64,34 @@ abstract contract RollupBase is
...
+        minimumAssertionPeriod = newPeriod;
+        emit ConfigurationChanged();
+    }
+
+    /// @inheritdoc IRollup
+    function setBaseStakeAmount(uint256 newAmount) public override onlyOwner {
```
saxenism, line 87: Change the function from public to external perhaps? That is the function signature in interface as well as the gas-efficient way to move forward assuming this function is not called inside the contracts.

PR #102, opened 2023-03-29
save more gas

easonnong:
# Saving more gas

Core changes:

- [[contracts/src/Rollup.sol#L532](https://github.com/SpecularL2/specular/blob/51a0826178f9f7389022b015c1a87023cb139b31/contracts/src/Rollup.sol#L532)]
- delete `i++`
- add `unchecked { i += 1; }`

Notes:

- change base on [this.](https://github.com/SpecularL2/specular/pull/95)
- at the end of the loop can save more gas by using `unchecked { i += 1; }`
- @LEAFERx

### DIFF
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -529,10 +529,11 @@ contract Rollup is RollupBase {
      */
     function countStakedZombies(uint256 assertionID) private view returns (uint256) {
         uint256 numStakedZombies = 0;
-        for (uint256 i = 0; i < zombies.length; i++) {
+        for (uint256 i = 0; i < zombies.length;) {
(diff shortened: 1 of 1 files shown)


saxenism:
Hey @easonnong, thanks for your contributions.

Looping in @ujvl to confirm whether gas optimization of our contracts is currently on our priority list or not.

Anyways, my suggestions would be 2-fold:
1. Create a list of all plausible gas optimizations in the Specular contracts, post that list in the PR description and then make changes. It would be much more coherent to review such a PR as compared to reviewing all gas optimizations in different PRs in a sort of irregular flow.
2. Regarding your `unchecked` maths optimization, it would be great if you could provide the list of reasonings behind **why you think an unchecked block is safe** for all the instances that you decide to introduce them at, in the PR description.

Apart from that, great work :)

PR #104, opened 2023-04-01
Wait for transaction confirmation in rollup services

LEAFERx:
# Goals of PR

Core changes:

- Sanitize check in SequencerInbox to avoid the wrong order of batches due to retry
- Wait for transaction confirmation in rollup services


saxenism (commented):
Yeah, so the smart contract changes look good to me and it should work.

However, speaking in terms of best practices, this feels quite hacky, is more susceptible to break moving forward and would be difficult to integrate for anything that uses Specular Contracts as a base layer.

Therefore my suggestion would be to introduce a 
```
struct Context{
    uint256 numTxns
    uint256 l2BlockNumber
    uint256 l2Timestamp
}
```
and pass an array of struct Context to the function `appendTxBatch` rather than assuming that the 2nd element would be the block number and the 3rd element would be the timestamp and so on. You'll simply do stuff like:

```
if (contexts[i].l2BlockNumber == latestL2BlockNumber + 1) {}
```

PR #105, opened 2023-04-06
Sequencer inbox `verifyTxInclusion`

saxenism:
# Goals of PR

* adds a test for `SequencerInbox.verifyTxInclusion`
* fixes the tx Data hashing in `SequencerInbox.appendTxBatch`

### DIFF
--- a/contracts/src/SequencerInbox.sol
+++ b/contracts/src/SequencerInbox.sol
@@ -22,7 +22,6 @@
 
 pragma solidity ^0.8.0;
 
-import "forge-std/console.sol";
 import "./ISequencerInbox.sol";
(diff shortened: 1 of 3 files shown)


contracts/src/SequencerInbox.sol
```diff
@@ -22,6 +22,7 @@
 
 pragma solidity ^0.8.0;
 
+import "forge-std/console.sol";
```
ujvl, line 25: can this be removed

ujvl (approved):
lgtm besides CI failing (not sure if it's just flakiness? or maybe just need to rm the import)

PR #107, opened 2023-04-14
[Draft] Refactor L2 node

ujvl:
Goals:
- Define clearer boundaries between Geth and Specular services (will make it easier to switch to Engine API later / implement other clients like Erigon).
- Refactor responsibilities of sequencer & validator (and enable running both simultaneously)
- Block-by-block derivation -- design heavily influenced by Optimism's bedrock implementation
- Re-org handling

Other minor changes:
- consolidated codepath for sequencer and deterministic sync
- decoupled services from Geth lifecycle
- configurability using TOML (overrides flags)
- removed dedicated indexer (run driver)
- replaced `sync.WaitGroup` usage w/ `errgroup.Group` to support goroutines returning errors

Note: a lot of the diff is just moving things around (e.g. containing Geth blockchain interaction in a single package)

TODO: 
- reintegrate challenge protocol
- assertion management


clients/geth/specular/cmd/utils/flags.go
```diff
@@ -129,39 +165,63 @@ func RegisterEthService(stack *node.Node, cfg *ethconfig.Config) (ethapi.Backend
...
+		L2Config: rollup.L2Config{
+			L2Endpoint:     "ws://0.0.0.0:4012", // TODO: read this from http params? or from a separate flag
+			L2ClefEndpoint: clefEndpoint,
+		},
+		SequencerConfig: rollup.SequencerConfig{
+			SequencerAccountAddr: sequencerAddr,
+			SequencerPassphrase:  sequencerPassphrase,
+			MinExecutionInterval: time.Duration(ctx.Uint64(RollupSequencerMinExecutionIntervalFlag.Name)) * time.Second,
+			MaxExecutionInterval: time.Duration(ctx.Uint64(RollupSequencerMaxExecutionIntervalFlag.Name)) * time.Second,
+			SequencingInterval:   time.Duration(ctx.Uint64(RollupSequencerSequencingIntervalFlag.Name)) * time.Second,
```
siosw, line 213: I ran into this while working with the refactored code. It seems like this ends up not using the default value.
`time.Duration(RollupSequencerSequencingIntervalFlag.Value) * time.Second` works for me.

ujvl: oh right good catch, i'll move the defaults into the flags

PR #109, opened 2023-04-17
Add Batch Validation

ayushguptacse14:
# Goals of PR

Core changes: Add Batch and Block Validation for L2

### DIFF
--- a/clients/geth/specular/rollup/services/base.go
+++ b/clients/geth/specular/rollup/services/base.go
@@ -238,6 +238,16 @@ func (b *BaseService) processTxBatchAppendedEvent(
 		return fmt.Errorf("Failed to split AppendTxBatch input into batches, err: %w", err)
 	}
 	log.Info("Decoded batch", "#txs", len(batch.Txs))
+	safeL2Head := b.Eth.BlockChain().CurrentBlock().Number().Uint64()
+	l1BlockHead, err := b.L1Client.BlockNumber(ctx)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


clients/geth/specular/rollup/services/base.go
```diff
@@ -251,6 +261,42 @@ func (b *BaseService) processTxBatchAppendedEvent(
...
+		isValid, validationResult := ValidateSequenceBlock(l2Head, l1BlockHead, sequenceBlock)
+		validationResults[i] = fmt.Sprintf("SequenceBlock %d: %s", i, validationResult)
+		if !isValid {
+			allValid = false
+		}
+	}
+	return allValid, validationResults
+}
+
+func ValidateSequenceBlock(l2Head uint64, l1BlockHead uint64, sequenceBlock *rollupTypes.SequenceBlock) (bool, string) {
```
LEAFERx, line 284: Use error instead of (bool, string)

clients/geth/specular/rollup/services/base.go
```diff
@@ -251,6 +261,42 @@ func (b *BaseService) processTxBatchAppendedEvent(
 	return nil
 }
 
+func ValidateTxBatch(l2Head uint64, l1BlockHead uint64, txBatch *rollupTypes.TxBatch) (bool, []string) {
```
LEAFERx, line 264: Use error instead of (bool, []string), just return if error occurs

LEAFERx (changes requested):

PR #111, opened 2023-04-18
Allow clef to be used as external signer

siosw:
# Goals of PR

Core changes:
- nodes can receive a clef endpoint to use for signing transactions

Running Clef & Nodes locally:
1. in the `sbin` directory, run `./configure.sh && ./clean.sh && ./init.sh`. the init script has been extended to also set up clef.
3. run `./start_clef.sh` to start a clef instanced that nodes can connect to
4. in sperate terminals, run a HH node, `USE_CLEF=true ./start_sequencer.sh` and `USE_CLEF=true ./start_validator.sh`

The clef ruleset is configured to auto approve all transactions to the rollup address.


clients/geth/specular/sbin/start_validator_clef.sh
```diff
@@ -0,0 +1,23 @@
...
+    --rollup.node 'validator' \
+    --rollup.coinbase=70997970c51812dc3a010c7d01b50e0d17dc79c8 \
+    --rollup.clefendpoint 'http://127.0.0.1:8550/' \
+    --rollup.l1endpoint 'ws://localhost:8545' \
+    --rollup.l1chainid 31337 \
+    --rollup.sequencer-addr '0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266' \
+    --rollup.sequencer-inbox-addr '0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512' \
+    --rollup.rollup-addr '0x5FC8d32690cc91D4c39d9d3abcBD16989F875707' \
+    --rollup.rollup-stake-amount 100 \
+    --rollup.l1-rollup-genesis-block 0
```
ujvl, line 23: lets introduce some sort of config file where you can read these values from (in every script) so that we don't need to hardcode them everywhere

one way to do it would be defining env variables like how `configure.sh` does (e.g. write a script called `configure_system.sh` or something

also if possible i'd just merge this with `start_sequencer.sh` and `start_validator.sh`, and pass in a flag to the scripts indicating to use clef (instead of creating separate scripts)

clients/geth/specular/data/ruleset.js
```diff
@@ -0,0 +1,5 @@
+function ApproveTx(r) {
```
ujvl, line 1: can you explain what this is for

siosw: With Clef we can specify a set of rules which determines which transactions should be approved or rejected. If no rules are provided for a TX, it must be manually approved through clef which would be impractical for us.

We can set this to approve all TXs but we probably want it to be more restrictive. The rules can be pretty expressive so we could think about doing stuff like rate limiting as well.

Docs for more details: https://geth.ethereum.org/docs/tools/clef/rules

clients/geth/specular/cmd/utils/flags.go
```diff
@@ -51,6 +51,11 @@ var (
 		Usage: "The sequencer/validator address to be unlocked (pass passphrash via --password)",
 		Value: "",
 	}
+	RollupClefEndpointFlag = &cli.StringFlag{
```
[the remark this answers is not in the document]
siosw: One Clef endpoint can be used for multiple addresses. We have to create one transactor per address like we do [here](https://github.com/SpecularL2/specular/blob/3eb0191011a618d08dc0ff6875251208c8b59736/clients/geth/specular/rollup/service.go#L48)

clients/geth/specular/Makefile
```diff
@@ -2,20 +2,25 @@
...
-install: bindings geth
+install: bindings geth clef
 
 contracts:
 	./sbin/compile_contracts.sh
 
 bindings: contracts
 	go generate ./...
 
-geth: bindings
```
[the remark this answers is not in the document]
siosw: you're right. I was thinking that there is a duplication [here](https://github.com/SpecularL2/specular/blob/3eb0191011a618d08dc0ff6875251208c8b59736/clients/geth/specular/Makefile#L5). But I guess we still want `make geth` to work on it's own.

I think we can remove `bindings` from line 5?

clients/geth/specular/sbin/start_validator_clef.sh
siosw: I added the config values to `configure.sh`, lmk if you agree with that, but I think it makes sense to have them all in one place.

starting nodes with clef can now be done like this: `USE_CLEF=true ./start_sequencer.sh`

clients/geth/specular/Makefile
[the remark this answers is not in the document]
ujvl: sure although i think `make` ignores already-built targets

clients/geth/specular/sbin/init.sh
```diff
@@ -6,3 +6,7 @@ cd $DATA_DIR
 $GETH_SPECULAR_DIR/build/bin/geth --datadir ./data_sequencer --networkid 13527 init ./genesis.json
 $GETH_SPECULAR_DIR/build/bin/geth --datadir ./data_validator --networkid 13527 init ./genesis.json
 $GETH_SPECULAR_DIR/build/bin/geth --datadir ./data_indexer --networkid 13527 init ./genesis.json
+
+cd $SBIN
+./init_clef.exp
```
ujvl, line 11: lets guard this with USE_CLEF as well (for this PR at least should preserve the legacy behavior by default)

clients/geth/specular/sbin/configure.sh
```diff
@@ -5,4 +5,17 @@ export GETH_SPECULAR_DIR=$SBIN/../
...
+export L1_ENDPOINT=ws://localhost:8545
+export L1_CHAIN_ID=31337
+export CLEF_ENDPOINT=http://127.0.0.1:8550
+
+export COINBASE_ADDR=f39fd6e51aad88f6f4ce6ab8827279cfffb92266
+export SEQUENCER_INBOX_ADDR=0xe7f1725E7734CE288F8367e1Bb143E90bb3F0512
+export SEQUENCER_ADDR=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
+export ROLLUP_ADDR=0x5FC8d32690cc91D4c39d9d3abcBD16989F875707
+
+export ROLLUP_STAKE_AMOUNT=100
```
ujvl, line 21: nit: hmm i'd separate this into a separate file since these params are quite different from the paths defined above

clients/geth/specular/sbin/init_clef.exp
```diff
@@ -0,0 +1,68 @@
...
+expect ">"
+send "ok\n"
+
+expect "Password: "
+send "$CLEF_PW\n"
+
+expect "Repeat password: "
+send "$CLEF_PW\n"
+
+# set account password for 0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
```
ujvl, line 20: we shouldn't hardcode these anywhere (even in comments) -- you can refer to them as "sequencer" etc instead

same goes for below and in `init_clef`

clients/geth/specular/data/ruleset.js
ujvl: got it can you just add a short comment at the top saying this is used in clef since that's not clear from the file name

also nit: can we use typescript here instead of js if possible

clients/geth/specular/sbin/init_clef.sh
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
+SBIN=`dirname $0`
+SBIN="`cd "$SBIN"; pwd`"
+. $SBIN/configure.sh
+. $SBIN/configure_systems.sh
```
ujvl, line 5:
```suggestion
. $SBIN/configure_system.sh
```

clients/geth/specular/data/ruleset.js
```diff
@@ -0,0 +1,22 @@
+// rules which govern clef key management behavior
+// for more info, see: https://geth.ethereum.org/docs/tools/clef/rules
+
+function ApproveTx(r) {
+  // all transactions not going to the rollup addr are rejected
+  // TODO: do we need additional / more fine grained rules?
+
+  // this file can't access env variables so addresses have to be hardcoded
```
ujvl, line 8: cant we just run `configure_system.sh` from within this script and then read the env variable

clients/geth/specular/sbin/init_clef.exp
```diff
@@ -0,0 +1,44 @@
...
+# attest ruleset hash
+expect ">"
+send "ok\n"
+
+expect "Password: "
+send "$CLEF_PW\n"
+
+interact
+
+
```
ujvl, line 44: nit: rm extra lines

ujvl (approved):

PR #112, opened 2023-04-25
Remove block# redundancy in appendTxBatch

siosw:
# Goals of PR

## Core changes:
In a `TxBatch` we no longer store the block number in every context. Instead we only store the first block number of the batch and can reconstruct the other block numbers by incrementing.

## Notes:
We can see this leads to a small reduction of gas cost

```
Blocks in Batch: 10
TXs per Block:   30

only first block number of the batch:
(runs: 10, μ: 1997645152, ~: 1997645110)

only first block number of the batch, context struct:
(runs: 10, μ: 1997787298, ~: 1997787299)

block number in every context:
(runs: 10, μ: 1997722666, ~: 1997722664)

```

```
Blocks in Batch: 5
TXs per Block:   30

only first block number of the batch:
(runs: 10, μ: 143393358, ~: 143393361)

only first block number of the batch, context struct:
(runs: 10, μ: 143414539, ~: 143414585)

block number in every context:
(runs: 10, μ: 143403939, ~: 143403944)

```

### DIFF
--- a/clients/geth/specular/rollup/client/l1client.go
+++ b/clients/geth/specular/rollup/client/l1client.go
@@ -32,7 +32,7 @@ type L1BridgeClient interface {
 	) event.Subscription
 	Close()
 	// ISequencerInbox.sol
-	AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, txBatch []byte) (*types.Transaction, error)
+	AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, firstL2Block *big.Int, txBatch []byte) (*types.Transaction, error)
 	WatchTxBatchAppended(opts *bind.WatchOpts, sink chan<- *bindings.ISequencerInboxTxBatchAppended) (event.Subscription, error)
 	FilterTxBatchAppendedEvents(opts *bind.FilterOpts) (*bindings.ISequencerInboxTxBatchAppendedIterator, error)
 	DecodeAppendTxBatchInput(tx *types.Transaction) ([]interface{}, error)
@@ -230,10 +230,10 @@ func (c *EthBridgeClient) Close() {
 	c.client.Close()
 }
 
-func (c *EthBridgeClient) AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, txBatch []byte) (*types.Transaction, error) {
+func (c *EthBridgeClient) AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, firstL2Block *big.Int, txBatch []byte) (*types.Transaction, error) {
 	c.mu.Lock()
 	defer c.mu.Unlock()
-	f := func() (*types.Transaction, error) { return c.inbox.AppendTxBatch(contexts, txLengths, txBatch) }
+	f := func() (*types.Transaction, error) { return c.inbox.AppendTxBatch(contexts, txLengths, firstL2Block, txBatch) }
 	return retryTransactingFunction(f, c.retryOpts)
 }
 
--- a/clients/geth/specular/rollup/client/l1client.go
+++ b/clients/geth/specular/rollup/client/l1client.go
@@ -32,7 +32,7 @@ type L1BridgeClient interface {
 	) event.Subscription
 	Close()
 	// ISequencerInbox.sol
-	AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, firstL2Block *big.Int, txBatch []byte) (*types.Transaction, error)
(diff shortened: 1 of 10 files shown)


contracts/src/SequencerInbox.sol
```diff
@@ -59,10 +59,12 @@ contract SequencerInbox is ISequencerInbox, Initializable, UUPSUpgradeable, Owna
     }
 
     /// @inheritdoc ISequencerInbox
-    function appendTxBatch(uint256[] calldata contexts, uint256[] calldata txLengths, bytes calldata txBatch)
-        external
-        override
-    {
+    function appendTxBatch(
+        uint256[] calldata contexts,
```
ujvl, line 63: how does the gas usage change if each context is a struct (instead of embedding the tuples in the vector directly)

clients/geth/specular/rollup/types/tx_batch.go
```diff
@@ -45,22 +45,26 @@ func NewTxBatch(blocks []*types.Block, maxBatchSize uint64) *TxBatch {
 	var contexts []SequenceContext
 	var txs []*types.Transaction
 	gasUsed := new(big.Int)
+
+	// TODO: do we need to handle the case where blocks is empty
```
ujvl, line 49: this is fine for now; it's not possible in the refactored version

clients/geth/specular/rollup/client/l1client.go
```diff
@@ -32,7 +32,7 @@ type L1BridgeClient interface {
 	) event.Subscription
 	Close()
 	// ISequencerInbox.sol
-	AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, txBatch []byte) (*types.Transaction, error)
+	AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, firstL2Block *big.Int, txBatch []byte) (*types.Transaction, error)
```
ujvl, line 35: nit for clarity (here and other places)
```suggestion
	AppendTxBatch(contexts []*big.Int, txLengths []*big.Int, firstL2BlockNumber *big.Int, txBatch []byte) (*types.Transaction, error)
```

clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -59,18 +59,36 @@ func (v *Validator) tryValidateAssertion(lastValidatedAssertion, assertion *roll
...
+		)
+
 		if currentBlockNum > currentChainHeight {
 			return errAssertionOverflowedLocalInbox
 		}
+
 		block = v.Chain().GetBlockByNumber(currentBlockNum)
+
 		if block == nil {
+			log.Warn("Could not get current block by number")
```
ujvl, line 84: I'd add this to the error msg directly by making `errAssertionOverflowedLocalInbox` into a struct and adding a Msg field which appends to the existing message in `Error()`

contracts/src/SequencerInbox.sol
siosw: I updated the gas data above with measurements for using a context struct.
Definitely a noticeable increase but idk if where to draw the line between optimizing gas cost / readability.
lmk what you think

ujvl: okay gotcha lets just continue with the current interface (no struct)

clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -26,9 +26,25 @@ type challengeCtx struct {
...
+
+func (e *errAssertionOverflowedLocalInbox) Is(tgt error) bool {
+	_, ok := tgt.(*errAssertionOverflowedLocalInbox)
+	return ok
+}
+
+
+
+
+
```
ujvl, line 47: nit: extraneous space (gofmt should fix this too)

clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -26,9 +26,25 @@ type challengeCtx struct {
...
+}
+
+func (e *errAssertionOverflowedLocalInbox) Error() string {
+	return fmt.Sprint("assertion overflowed local inbox with msg:", e.Msg)
+}
+
+func (e *errAssertionOverflowedLocalInbox) Is(tgt error) bool {
+	_, ok := tgt.(*errAssertionOverflowedLocalInbox)
+	return ok
+}
```
ujvl, line 42: oh do you actually need to explicitly implement this? i figured it works automatically since it's a struct

ujvl (approved):

PR #117, opened 2023-05-05
better predeploy genesis generation

siosw:
# Goals of PR

Core changes:

improve genesis configuration and set up predeploy UUPS proxy


contracts/src/pre-deploy/ERC1967Proxy.sol
```diff
@@ -0,0 +1,32 @@
+// SPDX-License-Identifier: MIT
```
LEAFERx, line 1: Don't copy-past this. Directly import from openzepppelin like this: https://github.com/SpecularL2/specular/pull/99/files#diff-1a8c10fea46dab63836b8dc63381912fd83e15ec4041a2d5bcb136db51e1d933R4

siosw: agree, I changed this file to just include the import statement. this forces `ERC1967Proxy.sol` to be compiled.

contracts/deploy/00_generate_configuration.ts
```diff
@@ -0,0 +1,40 @@
...
+const execPromise = util.promisify(exec);
+
+const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
+  // 1. generate genesis file
+  const dataPath = `${__dirname}/../../clients/geth/specular/data/`;
+  const baseGenesisPath = path.join(dataPath, "base_genesis.json");
+  const genesisPath = path.join(dataPath, "genesis.json");
+
+  // 2. init geth nodes
+  await execPromise(path.join(CLIENT_SBIN_DIR, "clean.sh"));
```
siosw, line 18: I'm not sure if we actually want to run this every time we start a node. But I found it helpful for testing purposes so far

contracts/deploy/00_generate_configuration.ts
```diff
@@ -0,0 +1,40 @@
...
+  await execPromise(path.join(CLIENT_SBIN_DIR, "init.sh"));
```
siosw, line 19: @LEAFERx I saw that you added node initialisation to `export_genesis.sh` in your refactor branch. I think we can go either way but to me it seems cleaner to keep them separate. lmk what you think

contracts/scripts/create_genesis.ts
```diff
@@ -82,14 +51,135 @@ const main = () => {
...
+// TODO: this is not very efficient,
+// we should be able to narrow down the search once we know
+// where the file will live / how it will be run
+async function getStorageLayout(contractName: string) {
+  // we can read the storage layout from the HH artifacts like this:
+  // 1. read the contracts debug file to get the build-info hash of the most recent compilation
+  // 2. get the full source path of the contract
+  // 3. use the full source path to look up the storage layout in the build-info file
+
+  const debugFile = await glob(`**/artifacts/**/${contractName}.dbg.json`, {
```
LEAFERx, line 81: the path is simply `contracts/artifacts`, but you shouldn't read json but directly access from `HardhatRuntimeEnvironment`

contracts/src/pre-deploy/ERC1967Proxy.sol
```diff
@@ -0,0 +1 @@
+import "@openzeppelin/contracts/proxy/ERC1967/ERC1967Proxy.sol";
```
LEAFERx, line 1: Don't do this, directly import the contract in js code

clients/geth/specular/data/base_genesis.json
```diff
@@ -13,15 +13,44 @@
   "difficulty": "1",
   "gasLimit": "8000000",
   "extradata": "",
-  "alloc": {
-    "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266": {
+  "preDeploy": [
+    {
+      "address": "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
```
LEAFERx, line 18: I'm not sure if this is better. To me using address as key helps to prevent duplicate pre-deploy on a single address when manually configuring

clients/geth/specular/data/base_genesis.json
```diff
@@ -13,15 +13,44 @@
...
+    {
+      "address": "0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc",
       "balance": "10000000000000000000"
+    },
+    {
+      "address": "0x0000000000000000000000000000000000000020",
+      "contract": "Faucet",
+      "balance": "100000000000000000000",
+      "storage": {
+        "_owner": "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
```
LEAFERx, line 34: although not sure if we want to implement in this pr (maybe a separate pr is better), but it would be helpful if we allow some "macro" here, e.g. use `<SEQUENCER-ADDR>` here instead of hard encoding can be much more easier for configuration.

contracts/deploy/00_generate_configuration.ts
LEAFERx: separate this out, move this script to somewhere outside `contracts` folder. putting the script here couples L1 and L2.

contracts/deploy/00_generate_configuration.ts
LEAFERx: it is just your last pr can't work without init so I temporarily fix it by calling`init.sh` to `export_genesis.sh`. In fact we shouldn't call these `init.sh` etc because they're working on the dev sequencer data folder. we should use another temp data folder at least.

contracts/deploy/01_deploy_SequencerInbox.ts
```diff
@@ -0,0 +1,37 @@
+import { HardhatRuntimeEnvironment } from "hardhat/types";
```
LEAFERx, line 1: move to `L1`, see #99

contracts/src/pre-deploy/ERC1967Proxy.sol
LEAFERx: Do this in js:
`import ERC1967Proxy from "@openzeppelin/contracts/build/contracts/ERC1967Proxy.json"; `

See `contracts/deploy/util.ts:4`

siosw: I changed the genesis generation so now if there is no hre compilation artifact, it looks in `node_modules/@openzeppelin`.

This comes with the limitation that we can't read the storage layout from contracts imported that way, which is fine in the case of the Proxy but something to keep in mind for the future.

contracts/deploy/00_generate_configuration.ts
siosw: I changed the export script. it creates a temporary data dir and deletes it upon completion

contracts/deploy/01_deploy_SequencerInbox.ts
siosw: Done

clients/geth/specular/sbin/export_genesis.sh
```diff
@@ -5,9 +5,14 @@ SBIN="`cd "$SBIN"; pwd`"
...
-    --datadir ./data_sequencer
+    --datadir ./tmp_data/
     --verbosity 0
 )
-$GETH_SPECULAR_DIR/build/bin/geth init --datadir ./data_sequencer ./base_genesis.json
+
+$GETH_SPECULAR_DIR/build/bin/geth init --datadir ./tmp_data ./genesis.json
 $GETH_SPECULAR_DIR/build/bin/geth "${args[@]}" dump 0
+
+rm -r tmp_data/
```
LEAFERx, line 18: `-rf` instead?

clients/geth/specular/data/base_genesis.json
LEAFERx: remember to remove this since we have a copy in `config/`

config/create_genesis.ts
```diff
@@ -0,0 +1,180 @@
...
+
+      const slot = ethers.utils.hexZeroPad(s[0], 32);
+      const value = ethers.utils.hexZeroPad(s[1], 32);
+      storage.set(slot, value);
+    }
+
+    data.set("storage", Object.fromEntries(storage));
+  }
+
+  if (alloc.has(p.address)) {
```
LEAFERx, line 168: check it earlier, e.g. at the beginning of the function

LEAFERx:
Can you also create a script to generate genesis from `config` folder to `clients/geth/specular/data` folder and update docker process e.g. `docker/scripts/geth.sh`?

clients/geth/specular/sbin/export_genesis.sh
siosw: i think it might be nice not to suppress errors for easier debugging. But I don't have strong opinions, any specific reason you suggested it, or just general preference?

contracts/src/pre-deploy/ERC1967Proxy.sol
siosw: yeah manually providing the storage slots will still work, just not resolving var names.

how do you think the versioning should work? only think I can think of rn is having the option to provide a contract hash in `base_genesis.json` and then check if that matches during genesis generation?

config/create_genesis.ts
```diff
@@ -0,0 +1,180 @@
...
+  contract: string | undefined;
+  storage: { [key: string]: string };
+};
+
+async function main() {
+  const { baseGenesisPath, genesisPath } = parseArgs();
+  await generateGenesisFile(baseGenesisPath, genesisPath);
+}
+
+export async function generateGenesisFile(
```
ujvl, line 22: nit: some brief documentation for each of these functions would be helpful

clients/geth/specular/data/base_genesis.json
```diff
@@ -13,15 +13,44 @@
   "difficulty": "1",
   "gasLimit": "8000000",
   "extradata": "",
-  "alloc": {
```
ujvl, line 16: whats the difference between the purpose of this file and the one in `config/`

ujvl: nvm just saw the other comment

config/create_genesis.ts
```diff
@@ -0,0 +1,180 @@
+import { BigNumber, ethers } from "ethers";
```
ujvl, line 1: hmm does it make sense to put this in this dir? (1) having all the scripts spread out across several directories makes it harder to figure out how to use the proj; and (2) i don't like mixing config files and scripts

ujvl:
do we need to update any READMEs

clients/geth/specular/sbin/export_genesis.sh
LEAFERx: yeah I think you're right

contracts/hardhat.config.ts
```diff
@@ -52,6 +52,8 @@ function createConfig(network: string) {
   //       they are used to interact with L1 contracts in L2 deploy scripts
   if (network === "specularLocalDev") {
     config.companionNetworks = { l1: "localhost" };
+    config.gas = 2100000;
```
LEAFERx, line 55: why this is needed?

siosw: without it, upgrading a proxy through the HH plugin fails with `code: 'UNPREDICTABLE_GAS_LIMIT'`

siosw: fixed, this is no longer required

config/create_genesis.ts
siosw: added some JSDOC to these functions

contracts/src/pre-deploy/UUPSPlaceholderV2.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-License-Identifier: Apache-2.0
```
LEAFERx, line 1: I don't think we need this

contracts/scripts/upgrade_placeholder.ts
```diff
@@ -0,0 +1,46 @@
...
+import { keccak256 } from "ethereumjs-util";
+import { ethers, upgrades } from "hardhat";
+
+dotenv.config({ path: __dirname + "/../.env" });
+
+const main = async () => {
+  const UUPSPlaceholderFactory = await ethers.getContractFactory(
+    "UUPSPlaceholder"
+  );
+  const V2Factory = await ethers.getContractFactory("UUPSPlaceholderV2");
```
LEAFERx, line 12: Don't use V2, just upgrade it to another placeholder implementation. this scripts is just for testing purpose. in real practice these placeholder should be upgraded to pre-deploys using deploy script under `deploy/l2`

config/hardhat.config.js
```diff
@@ -0,0 +1,8 @@
+/** @type import('hardhat/config').HardhatUserConfig */
```
LEAFERx, line 1: any reason of using js instead of ts?

siosw: changed to ts

contracts/scripts/upgrade_placeholder.ts
siosw: yeah sure, V2 was just for illustration / testing purposes. I removed it now

PR #120, opened 2023-05-24
Fix docker 

LEAFERx:
# Goals of PR

Core changes:

- Fix dockerfiles
- Move `.dockerignore` to correct place

Notes:

- Write notes here


siosw (approved):
LGTM

PR #124, opened 2023-06-02
Skip built targets in Makefile

ujvl:
- Improve build time by skipping already-built targets. We were using phony targets for everything before, which does not allow `make` to check changes in targets
- Ensure go-src is built (target `specular`). TODO in future PR: only build if go source is changed (i.e. make this a real target). 

This PR also updates bindings (should've happened in last contract PR)


clients/geth/specular/Makefile
```diff
@@ -1,26 +1,64 @@
-.PHONY: geth bindings install
+.PHONY: install specular clean geth-docker contracts
 
+GOSRC = ./...
 GOBIN = ./build/bin
+# TODO use env variables
```
siosw, line 5: maybe using env should be in this PR already, wdyt?

I just learned that we can use `include .env` in the Makefile to import the env variables ([docs](https://www.gnu.org/software/make/manual/html_node/Include.html))

ujvl: yeah but idt we can `include sbin/configure.sh`; we also can't run it from inside the Makefile since the variables won't be exported that way, so the only way is to probably convert `sbin/configure.sh` into a makefile

siosw: my idea would have been to move all env vars into an `.env` file so we can just include that in the Makefile.

`sbin/configure.sh` can just run `export $(cat .env | xargs)` to preserve the current functionality

LEAFERx: please move them to an env file. other changes lgtm. also consider remove binding changes

ujvl: to be fixed later

LEAFERx (approved):

PR #126, opened 2023-06-08
Avoid committing generated bindings

siosw:
# Goals of PR

Core changes:

1. add generated bindings to `.gitignore` & change the github actions accordingly
2. use the same Dockerfile for the geth & hardhat containers, this makes sense since they depend on each other.
3. use `docker/specular-datadir` as geth data directory whenever `docker compose` is used


clients/geth/specular/Makefile
```diff
@@ -20,7 +20,7 @@ clef:
 	@echo "Done building clef."
 	@echo "Run \"$(GOBIN)/clef\" to launch clef."
 
-geth-docker:
+geth-docker: bindings
```
ujvl, line 23: i think `geth-docker` and `geth` are equivalent now so we can just get rid of the former right

ujvl (approved):

LEAFERx (approved):

PR #129, opened 2023-06-22
Extraction of geth-specific methods of prover

Andy-Zhou2:
Extraction of geth-specific methods of prover

Core changes:

- Renaming of packages to remove naming conflicts (proof and proof/proof)
- Extracted out common interface to be used by all L2 execution layer clients, as listed at prover/state/api.go
- geth_prover package provides the wrappers around geth implementations of the interfaces


clients/geth/specular/prover/geth_prover/geth_evm_logger.go
```diff
@@ -0,0 +1,7 @@
+package geth_prover
+
+import "github.com/ethereum/go-ethereum/core/vm"
+
+type GethEVMLogger struct {
```
ujvl, line 5: is this used anywhere

clients/geth/specular/prover/state/api.go
```diff
@@ -0,0 +1,63 @@
...
+	Coinbase() common.Address
+	GasLimit() uint64
+	BlockNumber() *big.Int
+	Time() *big.Int
+	Difficulty() *big.Int
+	BaseFee() *big.Int
+	Random() *common.Hash
+}
+
+type SpecularEVMLoggerInterface interface {
```
ujvl, line 49: hm is this supposed to be empty

clients/geth/specular/prover/geth_prover/geth_evm_logger.go
Andy-Zhou2: No, will delete that

clients/geth/specular/prover/state/api.go
Andy-Zhou2: I can make L2ELClientConfig an interface instead of struct and then delete that

PR #133, opened 2023-07-12
New bridge UI

ayushguptacse14:
# Goals of PR

Core changes:

- Bridge UI updated to react

Areas to review:
- React Hooks
- Stepper View

Steps to Test locally:

1. Start L1 using ganache, hardhat doesn't support `eth_getProof` currently
npm install ganache --global
ganache --chain.chainId 31337 -b 2 -m "test test test test test test test test test test test junk"

2. run L1 migrations
npx hardhat deploy --tags L1 --network localhost

3. start sequencer
./start_sequencer.sh

4. run L2 migrations
npx hardhat deploy --tags L2 --network specularLocalDev

5. copy latest generates typechain-types (For master we can run copy_typechain_types.sh)

6. For imitating relayer (confirming oracle) run ./contracts/scripts/confirm_oracle.ts

7. For deploying ERC20 token run ./contracts/scripts/createERC20Token.ts and copy token address to token.ts


StefanVelkoski (commented):
Hey Ayush, good job on the PR.
It’s a big one with many changes. I have a couple of notices that should be addressed. 
I would be happy to hop on a call and re-create these issues.

UX:
  1. On refresh, it always takes you to the deposit page, and if you are on the withdrawal page, it asks you to switch networks which is annoying - can you please make it stay the same page after a refresh is clicked?
  2. Around every 30 seconds, a `Switch network` notification asks me to switch networks in Metamask - please update this; a network switch button should be 
  3. Withdrawal is not finalized - L2 transaction is created(you can access it in Blockscout), but the funds aren’t transferred to L1. Frontend also doesn’t change.
  4. The deposit works great, but after it is finalized, I have to refresh the page to make another deposit! Can we change this?

UI:
  1. Can we reduce the height and width of the main container and the fonts?
  2. Following our landing page navbar design, can we replace the central logo with just the Specular logo icon without the writing? https://specular.network/.
  3. Can we add the address, current network, and connect/disconnect tabs in the top right corner of the navbar?
  4. Please add some space between the navbar and the main container.
  5. Can we switch the footer design to the one in our Docs?  https://specular.network/docs/getting-started-user/
  6. Please update the font family and font sizes of the FAQs to follow the font of our Docs.

NIT:

  1. The 3rd FAQ has a bug when clicking the dropdown icon.
  2. On withdrawal, it should be ETH -> xDAI.


Notes:
Good job on the multiple-wallet integration!

bridge-frontend/.env
```diff
@@ -0,0 +1,2 @@
+REACT_APP_PROD=true
+SKIP_PREFLIGHT_CHECK=true
```
StefanVelkoski, line 2: we should add a linter

PR #135, opened 2023-07-18
Ch/challenge length

chalkandpaste:
# Goals of PR

Rewrite the `Challenge.sol` contract to have defender and challenger propose length of challenge and the final hash for the challenge, so that the pair which has the least length is chosen and bisection happens from there.

Core changes:

- Remove the `initializeChallengeLength` function in favor of `proposeChallenge`. `proposeChallenge` now takes proposals for the values of`numSteps` and `endStateHash` from both parties and chooses one such that the pair has the least `numSteps` of the two.
- Cleanup changes and add a few comments for readability.

Notes:

- Need to implement a time out for the challenge.

### DIFF
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -307,10 +307,11 @@ contract Rollup is RollupBase {
     }
 
     /// @inheritdoc IRollup
-    function challengeAssertion(
-        address[2] calldata players,
-        uint256[2] calldata assertionIDs
-    ) external override returns (address) {
+    function challengeAssertion( address[2] calldata players, uint256[2] calldata assertionIDs) 
+        external
+        override
+        returns (address)
+    {
 
         uint256 defenderAssertionID = assertionIDs[0];
         uint256 parentID = assertions[defenderAssertionID].parent;
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -351,8 +351,8 @@ contract Rollup is RollupBase {
             daProvider,
             IChallengeResultReceiver(address(this)),
             assertions[parentID].stateHash,
-            // TODO: confirm removing this. Let Defender & Challenger propose
-            /*assertions[defenderAssertionID].stateHash,*/
+            assertions[defenderAssertionID].stateHash,
+            assertions[challengerAssertionID].stateHash,
             challengePeriod
         );
         return challengeAddr;
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -155,6 +155,11 @@ contract Rollup is RollupBase {
     }
 
     /// @inheritdoc IRollup
+    function getLastConfirmedAssertionID() external view override returns (uint256) {
+        return lastConfirmedAssertionID;
+    }
+
+    /// @inheritdoc IRollup
     function isStakedOnAssertion(uint256 assertionID, address stakerAddress) external view override returns (bool) {
         return assertionState[assertionID].stakers[stakerAddress];
     }
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -307,12 +307,11 @@ contract Rollup is RollupBase {
     }
 
     /// @inheritdoc IRollup
-    function challengeAssertion( address[2] calldata players, uint256[2] calldata assertionIDs) 
+    function challengeAssertion(address[2] calldata players, uint256[2] calldata assertionIDs)
         external
         override
         returns (address)
     {
-
         uint256 defenderAssertionID = assertionIDs[0];
         uint256 parentID = assertions[defenderAssertionID].parent;
         uint256 challengerAssertionID = assertionIDs[1];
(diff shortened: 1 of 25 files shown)


contracts/src/Rollup.sol
```diff
@@ -314,22 +314,21 @@ contract Rollup is RollupBase {
...
+        }
+        if (challengerAssertionID > lastCreatedAssertionID) {
+            revert UnproposedAssertion();
+        }
+        if (lastConfirmedAssertionID >= defenderAssertionID) {
+            revert AssertionAlreadyResolved();
+        }
+        // Require that players have attested to sibling assertions.
+        if (parentID != assertions[challengerAssertionID].parent) {
+            revert NotSiblings();
```
ujvl, line 331: the scope definition here was made to get around another EVM quirk: there are limitations on the structure of the stack https://soliditydeveloper.com/stacktoodeep

this change is fine as long as it compiles

contracts/src/challenge/SymChallenge.sol
```diff
@@ -86,23 +86,46 @@ contract SymChallenge is ChallengeBase, ISymChallenge {
...
         if (bisectionHash != 0) {
             revert AlreadyInitialized();
         }
         require(_numSteps > 0, "INVALID_NUM_STEPS");
-        bisectionHash = ChallengeLib.initialBisectionHash(startStateHash, endStateHash, _numSteps);
-        // TODO: consider emitting a different event?
-        emit Bisected(bisectionHash, 0, _numSteps);
+
+        // Defender proposes `numSteps` and `endStateHash` first
+        if (turn == Turn.Defender && msg.sender == defender) {
```
ujvl, line 104: should we use the `onlyOnTurn` modifier in the function definition? then you can just check
`if (turn == Turn.Defender)` and also don't need to do the reverts below

contracts/src/challenge/SymChallenge.sol
```diff
@@ -86,23 +86,46 @@ contract SymChallenge is ChallengeBase, ISymChallenge {
...
 
         turn = Turn.Defender;
         lastMoveBlock = block.number;
         defenderTimeLeft = challengePeriod;
         challengerTimeLeft = challengePeriod;
     }
 
-    // TODO: use minimum of challenger/defender's proposed numSteps.
-    function initializeChallengeLength(uint256 _numSteps) external override onlyOnTurn {
+    function proposeChallenge(bytes32 _endStateHash, uint256 _numSteps) external {
```
ujvl, line 96: what happens in the special case where they both pass in the same hashes but for different `numSteps`:
```
// defender calls:
proposeChallenge(sameStateHash, 32)
// challenger calls:
proposeChallenge(sameStateHash, 64)
```

contracts/src/challenge/SymChallenge.sol
```diff
@@ -55,26 +56,25 @@ contract SymChallenge is ChallengeBase, ISymChallenge {
...
      */
+
     function initialize(
         address _defender,
         address _challenger,
         IVerifier _verifier,
         IDAProvider _daProvider,
         IChallengeResultReceiver _resultReceiver,
         bytes32 _startStateHash,
-        bytes32 _endStateHash,
```
ujvl, line 74: one issue here is a malicious defender can propose the correct hash—which isn't consistent with what the assertion committed to, but they nevertheless are able to win the challenge

contracts/src/challenge/SymChallenge.sol
```diff
@@ -149,6 +191,8 @@ contract SymChallenge is ChallengeBase, ISymChallenge {
         emit Bisected(bisectionHash, challengedSegmentStart, challengedSegmentLength);
     }
 
+    // TODO: If bisection is not executed because both have commited to the same has at the same `numStep`, verify
+    //        there are no additional ops on the stack after the last step.
```
ujvl, line 195: typo in `"same has"` (i assume you mean hash) + indentation looks off

ujvl (approved):

contracts/src/challenge/SymChallenge.sol
```diff
@@ -55,26 +58,30 @@ contract SymChallenge is ChallengeBase, ISymChallenge {
...
      * @param _verifier Address of the verifier contract.
      * @param _daProvider DA provider.
      * @param _resultReceiver Address of contract that will receive the outcome (via callback `completeChallenge`).
      * @param _startStateHash Commitment to agreed-upon start state.
-     * @param _endStateHash Commitment to disagreed-upon end state being challenged.
+     * @param _endStateDefenseHash Proposed DA end state hash of the defender.
+     * @param _endStateChallengeHash Proposed DA end state hash of the challenger.
+     * @param challengePeriod The duration that challenger/defender has for continuing the challenge
      */
+
```
ujvl, line 72: nit: extraneous newline between documentation and fn definition

chalkandpaste: Fixed

contracts/src/challenge/SymChallenge.sol
chalkandpaste: Fixed.

PR #142, opened 2023-08-15
add validator whitelist

MoonShiesty:
# Goals of PR

Core changes:

Add validator whitelist to prevent other actors from calling state-modifying functions.

I've done some basic testing, if this draft looks good, I'll add checks to validator functions and test.

### DIFF
--- a/contracts/deploy/l1/02_deploy_Rollup.ts
+++ b/contracts/deploy/l1/02_deploy_Rollup.ts
@@ -28,7 +28,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   }
 
   const { deployments, getNamedAccounts } = hre;
-  const { sequencer, deployer } = await getNamedAccounts();
+  const { sequencer, validator, deployer } = await getNamedAccounts();
   const sequencerInboxProxyAddress = (
     await deployments.get(getProxyName("SequencerInbox"))
   ).address;
@@ -46,6 +46,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     0, // uint256 _initialAssertionID
     0, // uint256 _initialInboxSize
     initialVMHash, // bytes32 _initialVMhash
+    [validator], // address[] calldata _validators
   ];
 
   await deployUUPSProxiedContract(hre, deployer, "Rollup", args);
--- a/contracts/src/IRollup.sol
+++ b/contracts/src/IRollup.sol
@@ -120,6 +120,12 @@ interface IRollup {
     /// @dev Thrown when there are zero stakers
     error NoStaker();
 
+    /// @dev Thrown when an attempt is made to grant a role to an account that already possesses the role.
+    error RoleAlreadyGranted();
+
+    /// @dev Thrown when an attempt is made to grant a role to an account that already possesses the role.
+    error NoRoleToRevoke();
+
     struct Staker {
         bool isStaked;
         uint256 amountStaked;
@@ -235,7 +241,34 @@ interface IRollup {
     // *** State mutation ***
 
     /**
+     * @notice Grant the validator role for the specified address.
+     *
+     * Emits: `ConfigChanged` event.
+     *
+     * @param validator Address to grant validator role.
+     */
+    function addValidator(address validator) external;
+
+    /**
+     * @notice Revoke the validator role for the specific address.
+     *
+     * Emits: `ConfigChanged` event.
+     *
+     * @param validator Address to revoke validator role.
+     */
+    function removeValidator(address validator) external;
+
+    /**
+     * @notice Revoke the senders validator role.
+     * @dev This function's purpose is to provide a mechanism for accounts to revoke their privileges if they are compromised.
+     *
+     * Emits: `ConfigChanged` event.
+     */
+    function removeOwnValidatorRole() external;
+
+    /**
      * @notice Deposits stake on staker's current assertion (or the last confirmed assertion if not currently staked).
+     * @notice Only callable by whitelisted validators.
      * @dev Currently uses Ether to stake; Must be > than defined threshold if this is a new stake.
      */
     function stake() external payable;
@@ -248,13 +281,15 @@ interface IRollup {
 
     /**
      * @notice Removes stakerAddress from the set of stakers and withdraws the full stake amount to stakerAddress.
-     * This can be called by anyone since it is currently necessary to keep the chain progressing.
+     * @notice Only callable by whitelisted validators.
+     * @dev This can be called by any whitelisted validator since it is currently necessary to keep the chain progressing.
      * @param stakerAddress Address of staker for which to unstake.
      */
     function removeStake(address stakerAddress) external;
 
     /**
      * @notice Advances msg.sender's existing stake to assertionID.
+     * @notice Only callable by whitelisted validators.
      * @param assertionID ID of assertion to advance stake to. Currently this must be a child of the current assertion.
      * TODO: generalize to arbitrary descendants.
      */
@@ -269,6 +304,7 @@ interface IRollup {
      * @notice Creates a new DA representing the rollup state after executing a block of transactions (sequenced in SequencerInbox).
      * Block is represented by all transactions in range [prevInboxSize, inboxSize]. The latest staked DA of the sender
      * is considered to be the predecessor. Moves sender stake onto the new DA.
+     * @notice Only callable by whitelisted validators.
      *
      * Emits: `AssertionCreated` and `StakerStaked` events.
      *
@@ -279,6 +315,7 @@ interface IRollup {
 
     /**
      * @notice Initiates a dispute between a defender and challenger on an unconfirmed DA.
+     * @notice Only callable by whitelisted validators.
      * @param players Defender (first) and challenger (second) addresses. Must be staked on DAs on different branches.
      * @param assertionIDs Assertion IDs of the players engaged in the challenge. The first ID should be the earlier-created and is the one being challenged.
      * @return Newly created challenge contract address.
@@ -289,6 +326,7 @@ interface IRollup {
 
     /**
      * @notice Confirms first unresolved assertion. Assertion is confirmed if and only if:
+     * @notice Only callable by whitelisted validators.
      * (1) there is at least one staker, and
      * (2) challenge period has passed, and
      * (3) predecessor has been confirmed, and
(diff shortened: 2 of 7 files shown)


contracts/src/IRollup.sol
```diff
@@ -220,6 +228,21 @@ interface IRollup {
      */
     function setBaseStakeAmount(uint256 newAmount) external;
 
+    /**
+     * @notice Set the addresses of the validator whitelist
```
chalkandpaste, line 232: Can you write a better notice? I'm thinking this interface can be a little unintuitive, in that if an address previously whitelisted, but doesn't have a state update supplied to this function, they will remain whitelisted (that is, they have to be specifically unwhitelisted). Maybe a better function name might clear this up as well? `updateValidators`.

MoonShiesty: pushed the change

MoonShiesty:
@chalkandpaste @siosw ready for review: added testing and renamed things for clarity.

I added the `whitelistOnly` modifier to the following validator functions which could be misused in testnet:

- `stake`
- `unstake`
- `removeStake`
- `advanceStake`
- `withdraw`
- `createAssertion`
- `challengeAssertion`
- `confirmFirstUnresolvedAssertion`
- `rejectFirstUnresolvedAssertion`

The way the whitelist is designed if we remove an already staked validator from the whitelist, that validator will not be able to `unstake` itself - another whitelisted addresses will have to call `removeStake` for them. They also won't be able to `withdraw` funds won in challenges. If this is an issue, we should remove the modifier for `unstake` and `withdraw`.

`unstake`/`advanceStake`/`createAssertion` already check that the caller is staked, but adding `whitelistOnly` allows us to disable all validator functionality for addresses that are already staked. This might be check depending on how we plan to use the whitelist.

chalkandpaste:
Re: `unstake` and `removeStake`... This could be solved by removing the stake when they are removed from the whitelist. Alternatively, but maybe not so clean, is to *not* use the modifier on `unstake` given it's restriction to the caller. It feels like maybe there's a third way... but nothing's coming to me.

Re: the second bit... this kinda indicates to me it might best to remove the stake when the validator is removed from the whitelist.

chalkandpaste:
Otherwise, everything looks good.

MoonShiesty:
> Re: `unstake` and `removeStake`... This could be solved by removing the stake when they are removed from the whitelist. Alternatively, but maybe not so clean, is to _not_ use the modifier on `unstake` given it's restriction to the caller. It feels like maybe there's a third way... but nothing's coming to me.
> 
> Re: the second bit... this kinda indicates to me it might best to remove the stake when the validator is removed from the whitelist.

This approach would be tricky because we don't want to remove the stake of a validator that is staked on an unconfirmed assertion.

I'm going to change the implementation so we don't `onlyWhitelist` `withdraw` - `withdraw` only affects the calling validator's state. This way un-whitelisted validators can remove earnings. 

I think we should still `onlyWhitelist` `unstake` because `removeStake` still allows whitelisted validators to return un-whitelisted stake. But I could go either way, as long we document.

I'm also going to document which functions might be whitelisted.

contracts/test/Rollup.t.sol
```diff
@@ -224,6 +228,130 @@ contract RollupTest is RollupBaseSetup {
         }
     }
 
+    ////////////////
+    // Admin
+    ///////////////
+
+    function test_initialize_validatorWhitelist_works() external {
```
siosw, line 235: very much a nit but i think the convention is to name it `_succeeds()`

contracts/src/Rollup.sol
```diff
@@ -210,7 +226,28 @@ contract Rollup is RollupBase {
     }
 
     /// @inheritdoc IRollup
-    function stake() external payable override {
+    function updateValidatorWhitelist(address[] calldata validatorWhitelistUpdate, bool[] calldata vals)
+        external
+        override
+        onlyOwner
```
siosw, line 232: should there be a way to for validators to revoke their own permissions? 
might be useful in the case of compromised keys

contracts/src/Rollup.sol
```diff
@@ -210,7 +226,28 @@ contract Rollup is RollupBase {
     }
 
     /// @inheritdoc IRollup
-    function stake() external payable override {
+    function updateValidatorWhitelist(address[] calldata validatorWhitelistUpdate, bool[] calldata vals)
```
siosw, line 229: my personal preference would be to split this into 2 functions like
```
function whitelistValidators(address[] calldata validatorsToAdd)
function unwhitelistValidators(address[] calldata validatorsToRemove)
```

contracts/package.json
```diff
@@ -42,8 +42,8 @@
     "typescript": "^4.9.3"
   },
   "dependencies": {
-    "@openzeppelin/contracts": "^4.8.3",
-    "@openzeppelin/contracts-upgradeable": "^4.8.3",
+    "@openzeppelin/contracts": "^4.9.3",
```
MoonShiesty, line 45: Updated to newer openzepplin with the AccessControlDefaultAdminRulesUpgradeable extension.

contracts/src/Rollup.sol
```diff
@@ -210,7 +229,53 @@ contract Rollup is RollupBase {
     }
 
     /// @inheritdoc IRollup
-    function stake() external payable override {
+    function addValidatorsToWhitelist(address[] calldata validators) external override onlyRole(DEFAULT_ADMIN_ROLE) {
+        for (uint256 i = 0; i < validators.length; i++) {
+            if (hasRole(VALIDATOR_WHITELIST_ROLE, validators[i])) {
```
MoonShiesty, line 234: We can simplify this check in the next OpenZeppelin version which returns a bool for grant/RevokeRole() to indicate whether the call succeeds.

contracts/test/Rollup.t.sol
```diff
@@ -224,11 +230,212 @@ contract RollupTest is RollupBaseSetup {
...
+            0, // initialAssertionID
+            0, // initialInboxSize
+            new address[](0) // validator whitelist
+        );
+
+        address[] memory validators = new address[](1);
+        validators[0] = alice;
+
+        // Adding a validator to whitelist as non-admin should fail
+        vm.expectRevert(
```
MoonShiesty, line 253: We can simplify this test with a selector in the next OpenZeppelin version which defines custom error.

contracts/src/Rollup.sol
MoonShiesty: good idea

MoonShiesty:
Ready for review. I changed the validator whitelist implementation to use role-based access control. There are currently two roles, `VALIDATOR_WHITELIST_ROLE` and `DEFAULT_ADMIN_ROLE`.

Role are managed by the admin role, which is initialized to the proxy deployer. The admin role is governed by `AccessControlDefaultAdminRules` which implements the following mitigations: 

- Only one account holds the `DEFAULT_ADMIN_ROLE` since deployment until it’s potentially renounced.

- Enforces a 2-step process to transfer the `DEFAULT_ADMIN_ROLE` to another account.

- Enforces a configurable delay between the two steps, with the ability to cancel before the transfer is accepted.

- The delay can be changed by scheduling, see [changeDefaultAdminDelay](https://docs.openzeppelin.com/contracts/4.x/api/access#AccessControlDefaultAdminRules-changeDefaultAdminDelay-uint48-).
- It is not possible to use another role to manage the `DEFAULT_ADMIN_ROLE`.

I set the initial delay to change the admin role to 3 days.

Also added more testing and documentation.

contracts/deploy/l1/02_deploy_Rollup.ts
```diff
@@ -46,6 +46,7 @@ const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
     0, // uint256 _initialAssertionID
     0, // uint256 _initialInboxSize
     initialVMHash, // bytes32 _initialVMhash
+    [validator], // address[] calldata _validators
```
LEAFERx, line 49: sequencer is also a validator

contracts/src/Rollup.sol
```diff
@@ -61,7 +67,6 @@ abstract contract RollupBase is
     }
 
     function __RollupBase_init() internal onlyInitializing {
-        __Ownable_init();
         __UUPSUpgradeable_init();
```
LEAFERx, line 70: do you need to init `AccessControlDefaultAdminRulesUpgradeable`?

contracts/src/Rollup.sol
```diff
@@ -115,6 +128,12 @@ contract Rollup is RollupBase {
         lastConfirmedAssertionID = _initialAssertionID;
         lastCreatedAssertionID = _initialAssertionID;
 
+        // Initialize role based access control
+        __AccessControlDefaultAdminRules_init(3 days, msg.sender);
```
LEAFERx, line 132: maybe move to `__RollupBase_init()` so we have a consistent place to initialize these contracts?

MoonShiesty:
I've left `unstake` as a non-whitelisted function, because there's no way for a validator to misuse the `unstake` function - they would only be wasting gas. Otherwise if we remove a validator, with active stake from the whitelist, they won't be able to withdraw their stake. They would need another a whitelisted validator to call `removeStake` on their behalf.

I think its better to design the contracts so that validators can always trust that their stake is available to withdraw after the staked assertion is confirmed.

contracts/src/Rollup.sol
```diff
@@ -126,20 +144,18 @@ contract Rollup is RollupBase {
             block.number // deadline (unchallengeable)
         );
         emit AssertionCreated(lastCreatedAssertionID, msg.sender, _initialVMhash);
-
-        __RollupBase_init();
```
MoonShiesty, line 130: I moved `__RollupBase_init` to the top of the function since we need to initialize the admin role, before we can add roles and initialize the initial validator whitelist.

MoonShiesty:
@LEAFERx thanks for the review, added the requested changes

LEAFERx (approved):

contracts/src/Rollup.sol
```diff
@@ -210,7 +229,53 @@ contract Rollup is RollupBase {
...
+
+    /// @inheritdoc IRollup
+    function removeValidatorsFromWhitelist(address[] calldata validators)
+        external
+        override
+        onlyRole(DEFAULT_ADMIN_ROLE)
+    {
+        for (uint256 i = 0; i < validators.length; i++) {
+            if (!hasRole(VALIDATOR_WHITELIST_ROLE, validators[i])) {
+                revert NoRoleToRevoke();
```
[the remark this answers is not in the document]
MoonShiesty: good point, especially since you'd need to look through events to see which accounts that have/have not been granted a role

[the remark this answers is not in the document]
MoonShiesty: done

MoonShiesty:
@chalkandpaste thanks for the review, I've changed `addValidator/removeValidator` to use a single address, rather than a list. Let me know if it looks good.

I also squashed down to a single commit.

chalkandpaste (approved):

ujvl (approved):

PR #146, opened 2023-08-25
Decouple sequencing, dissemination and validation

ujvl:
- Gets rid of obsolete sequencer code (but not all of it)
- Moves batch dissemination to its own service
- Moves validation (currently only assertion creation/confirmation) to its own service
- Updates config flags etc

There are multiple limitations in order to maintain compatibility but these will be resolved in the subsequent PR.


LEAFERx (changes requested):
please fix the node type issue causing validator to sync blocks with wrong coinbase. otherwise lgtm

LEAFERx (approved):

PR #147, opened 2023-08-25
Add Ability to Pause/Unpause Contracts

chalkandpaste:
# Goals of PR

Core changes:

- Implement `PausableUpgradeable` from OpenZeppelin for the SequencerInbox, Rollup and Bridge contracts
- Use modifier `whenPaused` to decorate `_authorizeUpgrade` so that upgrade can only be authorized when paused.
- Use modifier `whenNotPaused` to decorate particular functions that would progress the state of the contracts such that new consensus must be arrived at. Namely, prevent bridge transfers (requires confirmation), prevent new assertions from being created (requiring confirmation), preventing new stakers from staking and prevent new transaction batches from being appended (as they need to be asserted and confirmed),
- Testing some basic functionalities are correctly disabled when the contracts are paused.
- Adds a method `hasConsensus` for checking if `Rollup.sol` has reached a consensus state (all challenges have been resolved) which gets run for `_authorizeUpgrade()`.
- scripts which pause and check for consensus and which unpause the contracts for use when we need to upgrade

Notes:

- This isn't thoroughly unit tested, but I capture most of the important aspects (staking/unstaking/removingStake/progressing assertions/appendingTxs). In particular, we are missing tests for anything that doesn't use the `whenPaused`/`whenNotPaused`.
- Not decided on whether stakers can join or leave during pause (code currently allows for both).
- This does not test bridge code and overall there is a lack of testing bridge code, possibly because we don't have a good way of testing it? Something left to future work.
- Uncertain if upgrades should be blocked on consensus.

### DIFF
--- a/contracts/scripts/pause.ts
+++ b/contracts/scripts/pause.ts
@@ -35,7 +35,8 @@ async function pauseContract(
   const owner = new ethers.Wallet(privateKey, provider);
   const factory = await ethers.getContractFactory(contract, address);
   const c = factory.attach(address);
-  c.pause();
+  const tx = c.pause();
+  await tx.wait();
 }
 
 async function unpauseContract(
@@ -47,7 +48,8 @@ async function unpauseContract(
   const owner = new ethers.Wallet(privateKey, provider);
   const factory = await ethers.getContractFactory(contract, address);
   const c = factory.attach(address);
-  c.unpause();
+  const tx = c.unpause();
+  await tx.wait();
 }
 
 async function pauseContracts() {
--- a/contracts/scripts/pause.ts
+++ b/contracts/scripts/pause.ts
@@ -1,0 +1,86 @@
+import * as ethers from "ethers";
+
+const rollupAddr = process.env.ROLLUP_ADDR;
+const rollupPrKey = process.env.ROLLUP_PRIVATE_KEY;
+
+const sequencerAddr = process.env.SEQUENCER_ADDR;
+const sequencerPrKey = process.env.SEQUENCER_PRIVATE_KEY;
+
+const l1BridgeAddr = process.env.L1_BRIDGE_ADDR;
+const l1BridgePrKey = process.env.L1_BRIDGE_PRIVATE_KEY;
+
+const l1OracleAddr = process.env.L1_ORACLE_ADDR;
+const l1OraclePrKey = process.env.L1_ORACLE_PRIVATE_KEY;
+
+const l1PortalAddr = process.env.L1_PORTAL_ADDR;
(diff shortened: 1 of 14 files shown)


chalkandpaste:
Questions for @ujvl:

- Should we pause `unstake` and `removeStake` when we go to the pause state? These don't interfere with consensus, and as long as the upgrade hasn't taken place, it seems like a reasonable action to allow. I intuit there might be reasons for stopping, but I'm not 100% it's necessary.
- Alternatively, can we remove pausing from having new stakers join? New stakers get staked on the consensus assertion, so there is nothing additional to reach consensus.

contracts/scripts/pause.ts
```diff
@@ -0,0 +1,86 @@
...
+async function pauseContract(
+  contract: string,
+  address: string,
+  provider: ethers.providers.JsonRpcProvider,
+  privateKey: string
+) {
+  const owner = new ethers.Wallet(privateKey, provider);
+  const factory = await ethers.getContractFactory(contract, address);
+  const c = factory.attach(address);
+  c.pause();
```
LEAFERx, line 38: should it be something like `const tx = c.pause(); await tx.wait();`?

contracts/src/bridge/L1Oracle.sol
```diff
@@ -1,11 +1,11 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.4;
 
-import {Initializable} from "@openzeppelin/contracts-upgradeable/proxy/utils/Initializable.sol";
 import {UUPSUpgradeable} from "@openzeppelin/contracts-upgradeable/proxy/utils/UUPSUpgradeable.sol";
 import {OwnableUpgradeable} from "@openzeppelin/contracts-upgradeable/access/OwnableUpgradeable.sol";
+import {PausableUpgradeable} from "@openzeppelin/contracts-upgradeable/security/PausableUpgradeable.sol";
 
-contract L1Oracle is Initializable, UUPSUpgradeable, OwnableUpgradeable {
```
LEAFERx, line 8: why remove `Initializable`?

LEAFERx (approved):

contracts/src/bridge/L1Oracle.sol
chalkandpaste: `PausableUpgradeable` is already `Initializable`. I was having linearization errors in one spot (not here) and I thought it best to simplify since it's implicitly there. Can add back in here, I don't think it'll present compiler issues.

LEAFERx: sure just add it back so the entire codebase is consistent, all other contracts has `Initializable` with `PausableUpgradeable`

PR #151, opened 2023-09-01
bridge e2e testing

siosw:
# Goals of PR

Core changes:
- adds withdrawal tests to the existing bridge integration tests
- run the bridge integration tests as part of the e2e github action


e2e/sbin/test.sh
```diff
@@ -26,7 +26,13 @@ $SBIN_DIR/wait-for-it.sh -t 60 $HOST:$L2_HTTP_PORT
 # Run testing script
 cd $CONTRACTS_DIR
 npx ts-node scripts/testing.ts
-RESULT=$?
+exit_0 = $?
```
LEAFERx, line 29: we probably want to separate these tests so we know which succeeded and which failed. we can put them into different github action targets.

LEAFERx (approved):

LEAFERx:
lgtm

PR #179, opened 2023-10-05
validator should get inbox size from L1

siosw:
# Goals of PR

Core changes:

- Describe changes here

Notes:

- Write notes here

### DIFF
--- a/clients/geth/specular/rollup/services/validator/validator.go
+++ b/clients/geth/specular/rollup/services/validator/validator.go
@@ -7,9 +7,9 @@ import (
 	"math/big"
 	"time"
 
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/core/types"
-	"github.com/specularl2/specular/clients/geth/specular/rollup/rpc/bridge"
 	"github.com/specularl2/specular/clients/geth/specular/rollup/rpc/eth"
 	"github.com/specularl2/specular/clients/geth/specular/rollup/services/api"
 	"github.com/specularl2/specular/clients/geth/specular/utils/log"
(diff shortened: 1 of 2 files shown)


clients/geth/specular/rollup/services/validator/validator.go
```diff
@@ -180,7 +192,12 @@ func (v *Validator) getNextAssertionAttrs(ctx context.Context) (assertionAttribu
 	if err != nil {
 		return assertionAttributes{}, fmt.Errorf("failed to get finalized assertion attrs: %w", err)
 	}
-	return assertionAttributes{header.Number.Uint64(), header.Root}, nil
+
+	inboxSize, err := v.sequencerInbox.GetInboxSize(nil)
```
LEAFERx, line 196: this is incorrect, you need to get the inbox size corresponding to the header, or the header corresponding to the inbox size

config/data/base_genesis.json
```diff
@@ -26,6 +26,10 @@
       "address": "0x3c44cdddb6a900fa2b585dd299e03d12fa4293bc",
       "balance": "10000000000000000000"
     },
+    {
```
LEAFERx, line 29: what is this?

LEAFERx (changes requested):

clients/geth/specular/rollup/services/validator/validator.go
siosw: rolled back this change, inboxSize is now measured in terms of # blocks

config/data/base_genesis.json
siosw: added a new account for test scripts to use without changing the disseminator nonce

contracts/deploy/l2/01_deploy_L1Oracle.ts
```diff
@@ -5,7 +5,8 @@ import { deployUUPSProxiedContract } from "../utils";
 
 const func: DeployFunction = async function (hre: HardhatRuntimeEnvironment) {
   const { sequencer, deployer } = await hre.getNamedAccounts();
-  await deployUUPSProxiedContract(hre, deployer, "L1Oracle", [sequencer]);
+  const bridger = "0x8626f6940E2eb28930eFb4CeF49B2d1F2C9C1199"
+  await deployUUPSProxiedContract(hre, deployer, "L1Oracle", [bridger]);
```
siosw, line 9: this is a temporary fix to avoid changing the sequencer nonce during deployment. will no longer be needed once sequencer updates the oracle
(hardcoded address instead of putting it in env due to the temp nature of this fix)

PR #180, opened 2023-10-05
initial blob format contracts

MoonShiesty:
# Goals of PR

Create initial interface for batch data format

Core changes:

Change the signature of appendTxBatch

```
function appendTxBatch(    
        uint256 txBatchVersion,
        bytes calldata txBatch,
    ) external;
```
where:

```
txBatch = rlp([first_l2_block_number, blocks]), 

block = [timestamp, transaction_list]; blocks = [block1, block2, ...],
transaction_list = [tx1, tx2, ...].
```

Notes:

We could also encode `txBatchVersion` in the first byte of `txBatch` and remove the `txBatchVersion` argument.

```
txBatch = txBatchVersion ++ rlp([first_l2_block_number, blocks]), 

block = [timestamp, transaction_list]; blocks = [block1, block2, ...],
```

### DIFF
--- a/contracts/src/SequencerInbox.sol
+++ b/contracts/src/SequencerInbox.sol
@@ -74,9 +74,6 @@ contract SequencerInbox is ISequencerInbox, Initializable, UUPSUpgradeable, Owna
 
     /// @inheritdoc ISequencerInbox
     function appendTxBatch(
-        uint256[] calldata contexts,
-        uint256[] calldata txLengths,
-        uint256 firstL2BlockNumber,
         uint256 txBatchVersion,
         bytes calldata txBatch
     ) external override whenNotPaused {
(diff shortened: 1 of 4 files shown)


contracts/src/SequencerInbox.sol
```diff
@@ -88,46 +85,46 @@ contract SequencerInbox is ISequencerInbox, Initializable, UUPSUpgradeable, Owna
...
+        //     l2BlockNumber++;
+        // }
 
-        if (numTxs <= inboxSize) revert EmptyBatch();
-        uint256 start = inboxSize;
-        inboxSize = numTxs;
-        accumulators.push(runningAccumulator);
+        // if (numTxs <= inboxSize) revert EmptyBatch();
+        // uint256 start = inboxSize;
+        // inboxSize = numTxs;
```
chalkandpaste, line 124: This needs to *not* be commented out. Ref. issue solved in #179. Not sure what related code needs to also be uncommented, but the validators will need this.

PR #182, opened 2023-10-11
move hooks to lib directory

siosw:
- moves L1 fee hooks to a library, they will be imported in specular-geth (not implemented rn, still TODO)
- deletes unused entry point within sidecar


chalkandpaste (approved):
LGTM

PR #190, opened 2023-10-13
publish txBatch blob

MoonShiesty:
# Goals of PR

Create interface for batch data format, disseminate and derivate blocks using the new batch format.

Core changes:

Change the signature of appendTxBatch:

```
function appendTxBatch(    
        uint256 txBatchVersion,
        bytes calldata txBatch,
    ) external;
```

Modify the batch calldata to RLP encode transactions to the following format:
```
txBatch = rlp([first_l2_block_number, blocks]), 

block = [timestamp, transaction_list]; blocks = [block1, block2, ...],
transaction_list = [tx1, tx2, ...].
```

- Modify the sequencer to disseminate and derive RLP encoded batch data format.
- InboxSize increments by 1 each batch, instead of incrementing by the number of txs in a batch. This is because we're not currently doing any RLP decoding the SequencerInbox so we don't know the number of txs in a batch.
- Fix tests, some tests are commented out for now, some tests were no longer relevant and deleted, other tests needed to be changed to account for new InboxSize implementation.

Testing:

I tested by adding debug lines to confirm the batch format was properly RLP encoded (used this neat [tool](https://codechain-io.github.io/rlp-debugger)), and then confirmed the sequencer was able to decode and correctly derive submitted batches.

Submitted confirmed transactions via metamask.

### DIFF
--- a/clients/geth/specular/rollup/derivation/batch_builder.go
+++ b/clients/geth/specular/rollup/derivation/batch_builder.go
@@ -31,7 +31,7 @@ type BatchAttributes struct {
 }
 
 // TODO: find a better place to not hardcode this
-func TxBatchVersion() *big.Int { return big.NewInt(0) }
+func TxBatchVersion() byte { return byte(0x00) }
 
 type HeaderRef interface {
 	GetHash() common.Hash
--- a/clients/geth/specular/rollup/derivation/batch_builder.go
+++ b/clients/geth/specular/rollup/derivation/batch_builder.go
@@ -1,12 +1,12 @@
 package derivation
 
 import (
-	"bytes"
 	"fmt"
 	"io"
 	"math/big"
 
 	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/rlp"
 	"github.com/specularl2/specular/clients/geth/specular/rollup/types"
 	"github.com/specularl2/specular/clients/geth/specular/utils/log"
 )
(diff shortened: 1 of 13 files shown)


clients/geth/specular/rollup/derivation/batch_builder.go
```diff
@@ -20,19 +20,18 @@ type batchBuilder struct {
...
+}
+
 type BatchAttributes struct {
-	contexts           []*big.Int
-	txLengths          []*big.Int
-	firstL2BlockNumber *big.Int
-	txBatchVersion     *big.Int
-	txBatch            []byte // encoded batch of transactions.
+	FirstL2BlockNumber *big.Int
+	Blocks             []BlockAttributes
```
MoonShiesty, line 30: The RLP library uses reflection for encoding so I had to change these fields to public.

contracts/test/utils/RLPEncodedTransactions.sol
```diff
@@ -38,25 +38,24 @@ contract RLPEncodedTransactionsUtil is Test {
...
-    function _helper_sequencerInbox_appendTx_Version() internal returns (uint256) {
+    function _helper_sequencerInbox_appendTx_Version() internal pure returns (uint256) {
         // Update this as necessary
         return 0;
     }
 
-    function _helper_sequencerInbox_appendTx(uint256 numTxns) internal view returns (bytes memory, uint256[] memory) {
-        uint256[] memory transactionLengths = new uint256[](numTxns);
+    // TODO: RLP encode txBatch format
+    function _helper_sequencerInbox_appendTx(uint256 numTxns) internal view returns (bytes memory) {
```
MoonShiesty, line 47: I didn't updating this function to the new format because we don't check the contents of the batch calldata and the format is still subject to change.

If it makes sense to update this to the current version 0 format, I can add an RLPWriter library and encode correctly.

PR #195, opened 2023-10-17
clean up develop sbin

siosw:
# Goals of PR

make it easy to test the new system


Makefile
```diff
@@ -24,6 +24,7 @@ install: $(GETH_BIN_TARGET)
 geth: $(GETH_BIN_TARGET)
 
 sidecar: $(SIDECAR_BINDINGS_TARGET) $(shell find $(SIDECAR_DIR) -type f -name "*.go")
+	cd $(SIDECAR_DIR) && go generate ./...
```
chalkandpaste, line 27: No issue on this change to be clear, but a question maybe leading to further change: What does the $(SIDECAR_BINDINGS_TARGET) do then? I feel like this was the primary step for that target.

chalkandpaste (approved):

Makefile
siosw: the `$(SIDECAR_BINDINGS_TARGET) ` didn't generate the bindings when building sidecar. pushing a new change that keeps both steps separated to avoid unnecessary work.

btw what's the reasoning behind using bash vars as target names? i feel like the syntax is very confusing

PR #203, opened 2023-10-19
Missing dependencies for create genesis script

macwis:
# Goals of PR

Core changes:

- Missing dependencies to run `create_genesis.sh` out-of-the-box
- README update on the contracts to include `fondry install`
- (no change) done review if there are any unnecessary pnpm ts-node installs (which is not dev-dependency)

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,9 @@
   "devDependencies": {
     "@changesets/cli": "^2.26.2",
     "@pnpm/make-dedicated-lockfile": "^0.5.0",
-    "husky": "^8.0.1"
+    "@types/node": "^18.13.0",
+    "husky": "^8.0.1",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -12,6 +12,9 @@
   "devDependencies": {
     "@changesets/cli": "^2.26.2",
     "@pnpm/make-dedicated-lockfile": "^0.5.0",
-    "husky": "^8.0.1"
```
LEAFERx, line 15: why you remove `husky`

macwis: it's not removed, line 16 :) it's just got ordered differently after adding those missing packages

@ujvl speculated on discord about maybe actually removing it, but there is actually something in the pre-commit hook, so better leave it as it is.

ujvl (approved):

PR #204, opened 2023-10-19
publish txBatch blob

MoonShiesty:
# Goals of PR

Create interface for batch data format, disseminate and derivate blocks using the new batch format.

Core changes:

Change the signature of appendTxBatch:

```
function appendTxBatch(    
        bytes calldata txBatchData,
    ) external;
```

Modify the batch calldata to RLP encode transactions to the following format (batch format version 0, without epoch_num, epoch_hash):
```
########################
# L1 batcher tx format #
########################
batcher_tx_data = version || batch_data
# where...
# - version: uint8
# - data: bytes

#######################
# BatchData v0 format #
#######################
# For the first batcher tx in an epoch...
batch_data = 
	-	RLP([first_l2_block_num, batch_v0_list])
# - first_l2_block_num: uint256
# - batch_v0_list: List[batch_v0]
batch_v0 = [timestamp, tx_list]
# where...
# - timestamp: uint256
# - tx_list: bytes
```

- Modify the sequencer to disseminate and derive RLP encoded batch data format.
- InboxSize increments by 1 each batch, instead of incrementing by the number of txs in a batch. This is because we're not currently doing any RLP decoding the SequencerInbox so we don't know the number of txs in a batch.
- Fix tests, some tests are commented out for now, some tests were no longer relevant and deleted, other tests needed to be changed to account for new InboxSize implementation.

Testing:

I confirmed these changes work on the main branch, but can't test on develop without a working consensus client.

### DIFF
--- a/contracts/src/ISequencerInbox.sol
+++ b/contracts/src/ISequencerInbox.sol
@@ -28,34 +28,26 @@ import "./IDAProvider.sol";
  * @notice On-chain DA provider.
  */
 interface ISequencerInbox is IDAProvider {
-    event TxBatchAppended(uint256 batchNumber, uint256 startTxNumber, uint256 endTxNumber);
+    event TxBatchAppended(uint256 batchNumber);
 
-    /// @dev Thrown when the given tx inlcusion proof couldn't be verified.
+    /// @dev Thrown when the given tx inclusion proof couldn't be verified.
     error ProofVerificationFailed();
 
     /// @dev Thrown when sequencer tries to append an empty batch
     error EmptyBatch();
 
-    /// @dev Thrown when overflow occurs reading txBatch (likely due to malformed txLengths)
+    /// @dev Thrown when underflow occurs reading txBatch
+    error TxBatchDataUnderflow();
+
+    /// @dev Thrown when overflow occurs reading txBatch
     error TxBatchDataOverflow();
 
-    /// @dev Thrown when overflow occurs reading txBatch (likely due to malformed txLengths)
+    /// @dev Thrown when a transaction batch has an incorrect version
     error TxBatchVersionIncorrect();
 
     /**
      * @notice Appends a batch of transactions (stored in calldata) and emits a TxBatchAppended event.
-     * @param contexts Array of contexts, where each context is represented by a uint256 2-tuple:
-     * (numTxs, l2Timestamp). Each context corresponds to a single "L2 block".
-     * @param txLengths Array of lengths of each encoded tx in txBatch.
-     * @param firstL2BlockNumber The block number of the first "L2 block" included in this batch.
-     * @param txBatchVersion The serialization version of the submitted tx batch
-     * @param txBatch Batch of RLP-encoded transactions.
+     * @param txBatchData Batch of RLP-encoded transactions.
      */
-    function appendTxBatch(
-        uint256[] calldata contexts,
-        uint256[] calldata txLengths,
-        uint256 firstL2BlockNumber,
-        uint256 txBatchVersion,
-        bytes calldata txBatch
-    ) external;
+    function appendTxBatch(bytes calldata txBatchData) external;
 }
--- a/contracts/src/ISequencerInbox.sol
+++ b/contracts/src/ISequencerInbox.sol
@@ -28,7 +28,7 @@ import "./IDAProvider.sol";
  * @notice On-chain DA provider.
  */
 interface ISequencerInbox is IDAProvider {
-    event TxBatchAppended(uint256 batchNumber);
+    event TxBatchAppended();
 
     /// @dev Thrown when the given tx inclusion proof couldn't be verified.
     error ProofVerificationFailed();
--- a/contracts/src/ISequencerInbox.sol
+++ b/contracts/src/ISequencerInbox.sol
@@ -36,10 +36,10 @@ interface ISequencerInbox is IDAProvider {
     /// @dev Thrown when sequencer tries to append an empty batch
     error EmptyBatch();
 
-    /// @dev Thrown when underflow occurs reading txBatch
+    /// @dev Thrown when underflow occurs reading txBatchData
     error TxBatchDataUnderflow();
 
-    /// @dev Thrown when overflow occurs reading txBatch
+    /// @dev Thrown when overflow occurs reading txBatchData
     error TxBatchDataOverflow();
 
     /// @dev Thrown when a transaction batch has an incorrect version
--- a/contracts/src/ISequencerInbox.sol
+++ b/contracts/src/ISequencerInbox.sol
@@ -47,7 +47,21 @@ interface ISequencerInbox is IDAProvider {
 
     /**
      * @notice Appends a batch of transactions (stored in calldata) and emits a TxBatchAppended event.
-     * @param txBatchData Batch of RLP-encoded transactions.
+     * @param txBatchData Batch of RLP-encoded transactions, encoded as:
+     * txBatchData format:
+     *   txBatchData = version || batchData (|| is concatenation)
+     *   where:
+     *   - version: uint8
+     *   - data: bytes
+     * batchData format:
+     *   batchData = RLP([firstL2BlockNum, batchList])
+     *   where:
+     *   - firstL2BlockNum: uint256
+     *   - batchList: List[blockData]
+     *   blockData = [timestamp, txList]
+     *   where:
+     *   - timestamp: uint256
+     *   - txList: bytes
      */
     function appendTxBatch(bytes calldata txBatchData) external;
 }
--- a/contracts/test/SequencerInbox.t.sol
+++ b/contracts/test/SequencerInbox.t.sol
@@ -94,9 +94,10 @@ contract SequencerInboxTest is SequencerBaseSetup {
         seqIn.appendTxBatch(hex"00");
     }
 
-    function test_appendTxBatch_emptyBatch_succeeds() public {
+    function test_appendTxBatch_invalidVersion_reverts() public {
+        vm.expectRevert(ISequencerInbox.TxBatchVersionIncorrect.selector);
         vm.prank(sequencerAddress);
-        seqIn.appendTxBatch(hex"00");
+        seqIn.appendTxBatch(hex"01"); // version 0 is the only valid version
     }
 
     //////////////////////////////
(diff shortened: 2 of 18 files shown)


contracts/src/ISequencerInbox.sol
```diff
@@ -28,34 +28,26 @@ import "./IDAProvider.sol";
...
 
     /**
      * @notice Appends a batch of transactions (stored in calldata) and emits a TxBatchAppended event.
-     * @param contexts Array of contexts, where each context is represented by a uint256 2-tuple:
-     * (numTxs, l2Timestamp). Each context corresponds to a single "L2 block".
-     * @param txLengths Array of lengths of each encoded tx in txBatch.
-     * @param firstL2BlockNumber The block number of the first "L2 block" included in this batch.
-     * @param txBatchVersion The serialization version of the submitted tx batch
-     * @param txBatch Batch of RLP-encoded transactions.
+     * @param txBatchData Batch of RLP-encoded transactions.
```
ujvl, line 50: can you add the full format here

services/sidecar/rollup/services/disseminator/batch_disseminator.go
```diff
@@ -190,14 +191,17 @@ func (d *BatchDisseminator) disseminateBatch(ctx context.Context) error {
...
-	)
+	rlpBatchData, err := rlp.EncodeToBytes(batchAttrs)
+	if err != nil {
+		return fmt.Errorf("failed to encode batch: %w", err)
+	}
+
+	// Prepend batch data with version
+	txBatchData := make([]byte, len(rlpBatchData)+1)
+	txBatchData[0] = derivation.TxBatchVersion()
+	copy(txBatchData[1:], rlpBatchData)
```
ujvl, line 202: can we use `rlp.Encode` and pass in a buffer to avoid the copy?

contracts/src/ISequencerInbox.sol
MoonShiesty: done

services/sidecar/rollup/services/disseminator/batch_disseminator.go
MoonShiesty: fixed

services/sidecar/rollup/services/disseminator/interface.go
```diff
@@ -21,19 +21,15 @@ type BuildPayloadResponse = engine.ForkChoiceResponse
...
 
 type TxManager interface {
 	AppendTxBatch(
 		ctx context.Context,
-		contexts []*big.Int,
-		txLengths []*big.Int,
-		firstL2BlockNumber *big.Int,
-		txBatchVersion *big.Int,
-		txBatch []byte,
+		batchData *[]byte,
```
LEAFERx, line 32: what is `*[]byte`? `[]byte` is already a reference type.

services/sidecar/rollup/derivation/derivation_block.go
```diff
@@ -48,41 +49,37 @@ func (e *DecodeTxBatchError) Error() string {
...
 	}
-	var (
-		batchOffset, numTxs uint64
-		blocks              = make([]DerivationBlock, 0, len(contexts)/2)
-		blockNumber         = firstL2BlockNumber.Uint64()
-	)
-	for i := 0; i < len(contexts); i += 2 {
+
+	txBatchVersion := txBatchData[0]
+	if txBatchVersion != TxBatchVersion() {
```
LEAFERx, line 64: can we dispatch to different version of decoders instead of asserting 1 version? For other version return unimplemented.

services/sidecar/rollup/derivation/derivation_block.go
```diff
@@ -48,41 +49,37 @@ func (e *DecodeTxBatchError) Error() string {
...
+		return nil, &DecodeTxBatchError{fmt.Sprintf("invalid tx batch version")}
+	}
+
+	var decodedBatch BatchAttributes
+	if err := rlp.Decode(bytes.NewReader(txBatchData[:1]), &decodedBatch); err != nil {
```
LEAFERx, line 69: should be `txBatchData[1:]`?

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -7,32 +7,32 @@ import (
...
 	"github.com/specularL2/specular/services/sidecar/utils/log"
 )
 
 type batchBuilder struct {
 	maxBatchSize uint64
 
-	pendingBlocks   []DerivationBlock
-	builtBatchAttrs *BatchAttributes
+	pendingBlocks  []DerivationBlock
+	builtBatchData *[]byte
```
LEAFERx, line 19: dont do `*[]byte`, just use `[]byte`.

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -81,52 +81,62 @@ func (b *batchBuilder) Build() (*BatchAttributes, error) {
...
 		}
 	}
 	// Construct batch attributes.
-	var (
-		firstL2BlockNumber = big.NewInt(0).SetUint64(b.pendingBlocks[0].BlockNumber())
-		txBatchVersion     = big.NewInt(0) // TODO: Find a better place to not hard code this
-		attrs              = &BatchAttributes{contexts, txLengths, firstL2BlockNumber, txBatchVersion, buf.Bytes()}
-	)
+	attrs := &BatchAttributes{firstL2BlockNumber, blocks}
 	log.Info("Serialized l2 blocks", "first", firstL2BlockNumber, "last", b.pendingBlocks[idx].BlockNumber())
```
LEAFERx, line 125: is it possible that `idx` overflows `pendingBlocks` when the previous loop reads all pending blocks?

LEAFERx (changes requested):
I didn't look into solidity tests. I assume these tests would pass, right?

services/sidecar/rollup/derivation/batch_builder.go
MoonShiesty: coming from C++ i've been confused about whether to pass large variables by "value" or "pointer".

just did some research and seems like the internet's consensus is that parameters in go should only be "passed-by-pointer" if you wish to allow the function to modify the value passed to it. Because, escape analysis allows go to optimize "pass-by-value". does this sound right?

services/sidecar/rollup/derivation/batch_builder.go
MoonShiesty: it's not possible, this loop:
```for idx, block = range b.pendingBlocks {```
terminates with `idx` equal to the index of last item in the array, and `block` equal to the last item in the array.

its probably cleaner to rewrite as:

```log.Info("Serialized l2 blocks", "first", firstL2BlockNumber, "last", block.Number())```

Comment (LEAFERx) [reply services/sidecar/rollup/derivation/batch_builder.go:125]: what about the next few lines where there's a `b.pendingBlocks[idx + 1]`?

Comment (MoonShiesty) [reply services/sidecar/rollup/derivation/batch_builder.go:125]: `b.pendingBlocks[idx + 1]` will correctly truncate the slice to an empty slice if all blocks are processed however:

 when `b.pendingBlocks` is empty, the loop will not execute and `idx` will not be assigned a value. When the line `b.pendingBlocks` = `b.pendingBlocks[idx+1:]` executes, it would panic due to an index out of range error.

this shouldn't happen because we don't call `serializeToAttrs` unless there is at least one pending block:

https://github.com/SpecularL2/specular/blob/d8b2a27b619e1dcbc37c3a535558f122415f7041/services/sidecar/rollup/derivation/batch_builder.go#L77

but i will cleanup this code to make it less likely future changes unknowingly break things.

File: services/sidecar/rollup/derivation/batch_builder.go [hunk 8]
```diff
@@ -81,52 +82,66 @@ func (b *batchBuilder) Build() (*BatchAttributes, error) {
...
 }
 
 func (b *batchBuilder) Advance() {
-	b.builtBatchAttrs = nil
+	b.builtBatchData = nil
 }
 
 func (b *batchBuilder) serializeToAttrs() (*BatchAttributes, error) {
+	if len(b.pendingBlocks) == 0 {
+		return nil, errors.New("no pending blocks")
```
LEAFERx, line 100: we shouldn't return an error here? this will terminate the entire building process i guess?

MoonShiesty: i don't think returning an empty batch here makes sense since there's not value to populate `FirstL2BlockNumber`

contracts/test/SequencerInbox.t.sol
```diff
@@ -91,19 +91,12 @@ contract SequencerInboxTest is SequencerBaseSetup {
...
-        uint256[] memory contexts = new uint256[](1);
-        uint256[] memory txLengths = new uint256[](1);
-        uint256 txBatchVersion = _helper_sequencerInbox_appendTx_Version();
-        seqIn.appendTxBatch(contexts, txLengths, 1, txBatchVersion, "0x");
+        seqIn.appendTxBatch(hex"00");
     }
 
-    function test_appendTxBatch_emptyBatch_reverts() public {
-        vm.expectRevert(ISequencerInbox.EmptyBatch.selector);
+    function test_appendTxBatch_emptyBatch_succeeds() public {
```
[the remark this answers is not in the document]
MoonShiesty: done

services/sidecar/rollup/derivation/derivation_block.go
MoonShiesty: added a switch statement to dispatch, right now the only decoder is `blocksFromV0Data`

services/sidecar/rollup/derivation/derivation_block.go
MoonShiesty: fixed

MoonShiesty:
> I didn't look into solidity tests. I assume these tests would pass, right?

test are all passing

@LEAFERx made the suggested changes, and rebased

services/sidecar/rollup/derivation/derivation_block.go
```diff
@@ -48,41 +49,37 @@ func (e *DecodeTxBatchError) Error() string {
 
 // Decodes the input of `SequencerInbox.appendTxBatch` call
 func BlocksFromData(calldata []any) ([]DerivationBlock, error) {
```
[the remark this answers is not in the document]
LEAFERx: `calldata[0]` is `[]byte` but `calldata` itself is hidden under reflection

ujvl (approved):

LEAFERx (approved):

PR #206, opened 2023-10-20
Frontend separated for Blockscout

ayushguptacse14:
# Goals of PR

Core changes:

- Separate frontend for blockscout


macwis (approved):
Tested locally, works just fine with Docker :)

PR #207, opened 2023-10-20
sidecar docker

siosw:
# Goals of PR
- containerise sidecar
- ideally have compose files to spin up all components of the network

### DIFF
--- a/docker/scripts/sidecar.sh
+++ b/docker/scripts/sidecar.sh
@@ -1,0 +1,13 @@
+#!/bin/sh
+
+sidecar \
+    --rollup.sequencer \
+    --rollup.l1.endpoint $L1_ENDPOINT \
+    --rollup.l1.chainid $L1_CHAIN_ID \
+    --rollup.l1.sequencer-inbox-addr $SEQUENCER_INBOX_ADDR \
+    --rollup.l1.rollup-addr $ROLLUP_ADDR \
+    --rollup.l1.rollup-genesis-block $GENESIS_L1_BLOCK_NUM \
+    --rollup.l2.chainid $NETWORK_ID \
+    --rollup.sequencer.addr $SEQUENCER_ADDR \
+    --rollup.validator \
+    --rollup.validator.addr $VALIDATOR_ADDR \
--- a/docker/services/docker-compose-sidecar.yml
+++ b/docker/services/docker-compose-sidecar.yml
@@ -1,0 +1,11 @@
+services:
+  sidecar:
+    image: ${L2_DOCKER_TAG:-specularl2/sidecar:latest}
+    build:
+      context: ../..
+      dockerfile: ./services/sidecar/Dockerfile
+    volumes:
+      - ${SPECULAR_DATADIR:-$PWD/data}:/specular
+    env_file:
+      - ../../e2e/data/e2e.env
+    entrypoint: sidecar.sh
--- a/services/sidecar/cmd/sidecar/main.go
+++ b/services/sidecar/cmd/sidecar/main.go
@@ -2,11 +2,10 @@ package main
 
 import (
 	"context"
-	"bytes"
 	"math"
 	"math/big"
-	"reflect"
 	"os"
+	"reflect"
 
 	"github.com/urfave/cli/v2"
 	"golang.org/x/sync/errgroup"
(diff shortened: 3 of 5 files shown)


docker/services/docker-compose-sidecar.yml
```diff
@@ -0,0 +1,11 @@
+services:
```
MoonShiesty, line 1: docker-compose-l1geth.yml starts with a version, do we need that here as well?

services/sidecar/cmd/sidecar/main.go
```diff
@@ -132,10 +103,8 @@ func createTxManager(
 
 // Creates a transactor for the given account address, either using the clef endpoint or passphrase.
 func createTransactor(
-	mgr accountManager,
 	accountAddress common.Address,
 	clefEndpoint string,
-	passphrase string,
```
ujvl, line 138: this should still be passed in so you're not tied to a specific key

docker/scripts/sidecar.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/sh
```
ujvl, line 1: can we just use the existing `sbin` script?

docker/scripts/sidecar.sh
```diff
@@ -0,0 +1,13 @@
...
+    --rollup.sequencer \
+    --rollup.l1.endpoint $L1_ENDPOINT \
+    --rollup.l1.chainid $L1_CHAIN_ID \
+    --rollup.l1.sequencer-inbox-addr $SEQUENCER_INBOX_ADDR \
+    --rollup.l1.rollup-addr $ROLLUP_ADDR \
+    --rollup.l1.rollup-genesis-block $GENESIS_L1_BLOCK_NUM \
+    --rollup.l2.chainid $NETWORK_ID \
+    --rollup.sequencer.addr $SEQUENCER_ADDR \
+    --rollup.validator \
+    --rollup.validator.addr $VALIDATOR_ADDR \
```
ujvl, line 13: extra `\` at the end -- does this run?

docker/services/docker-compose-sidecar.yml
```diff
@@ -0,0 +1,11 @@
...
+  sidecar:
+    image: ${L2_DOCKER_TAG:-specularl2/sidecar:latest}
+    build:
+      context: ../..
+      dockerfile: ./services/sidecar/Dockerfile
+    volumes:
+      - ${SPECULAR_DATADIR:-$PWD/data}:/specular
+    env_file:
+      - ../../e2e/data/e2e.env
```
ujvl, line 10: does the env file path need to be relative to the context defined above?

docker/services/docker-compose-l2geth.yml
```diff
@@ -1,13 +0,0 @@
...
-  l2geth:
-    image: ${L2_DOCKER_TAG:-specularl2/l2geth:latest}
-    build:
-      context: ../..
-      dockerfile: ./clients/geth/specular/Dockerfile
-    volumes:
-      - ${SPECULAR_DATADIR:-$PWD/specular-datadir}:/specular
-    env_file:
-      - ../envs/common.env
-    entrypoint: geth.sh
```
ujvl, line 13: can we delete the script too? (if we delete both scripts can delete the whole dir)

docker/scripts/sidecar.sh
siosw: this runs but removed it for clarity

docker/services/docker-compose-sidecar.yml
siosw: no, relative to the file. it throws an error if the file is not found so I'm sure it's not failing silently

docker/scripts/sidecar.sh
siosw: agree that this would be nice but the sbin script relies on a `configure.sh` script that isn't easily ported to the docker image without preserving the entire directory structure of the mono repo.
I think we should keep it for now and try to consolidate them when we have a bit more time

docker/services/docker-compose-l2geth.yml
siosw: deleting this for now since it's not being used

PR #216, opened 2023-10-27
use l1 safe block for l2 genesis timestamp

MoonShiesty:
# Goals of PR

This PR uses the L1 safe block, as the timestamp for the L2 genesis file.

after this change, `genesis.json` should have the following field, with a non-zero value:

```"timestamp": 1698446636```

now create_genesis.ts expects an `--l1network` flag which is used to fetch the latest safe block. `create_genesis.sh` creates and destroys an L1, if one isn't already running for `create_genesis.ts`

I tested to confirm the new genesis.json file has a similar timestamp to the latest l1 safe block

```
services/el_clients/go-ethereum/build/bin/geth attach -exec "eth.getBlockByNumber('latest').timestamp" ws://172.17.0.1:8545
```

### DIFF
--- a/config/src/create_genesis.ts
+++ b/config/src/create_genesis.ts
@@ -19,10 +19,12 @@ async function main() {
   const baseGenesisPath = parseFlag("--in")
   const defaultGenesisPath = path.join(path.dirname(baseGenesisPath), "genesis.json");
   const genesisPath = parseFlag("--out", defaultGenesisPath)
-  await generateGenesisFile(baseGenesisPath, genesisPath);
+  const l1Network = parseFlag("--l1network")
+  await generateGenesisFile(l1Network, baseGenesisPath, genesisPath);
 }
 
 export async function generateGenesisFile(
+  l1Network: string,
   baseGenesisPath: string,
   genesisPath: string
 ) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: config/src/create_genesis.ts [hunk 1]
```diff
@@ -36,6 +38,15 @@ export async function generateGenesisFile(
   baseGenesis.preDeploy = undefined;
   baseGenesis.alloc = Object.fromEntries(alloc);
 
+  try {
+    const provider = new ethers.providers.WebSocketProvider(l1Network)
+    const block = await provider.getBlock("safe")
+    baseGenesis.timestamp = block.timestamp
+    provider._websocket.terminate();
```
MoonShiesty, line 45: ethers v6 provides a better handle to close the websocket, but we need to do this in v5 so the script terminates https://github.com/ethers-io/ethers.js/issues/1053

sbin/create_genesis.sh
```diff
@@ -25,8 +25,49 @@ relpath () {
...
+    else
+        echo "invalid value for L1_STACK: $L1_STACK"
+        exit 1
+    fi
+else
+    L1_PID=0
+    echo "L1 already started..."
+fi
+
+
```
MoonShiesty, line 63: I duplicated a lot of code from `start_l1.sh` but I didn't want to create an entire docker process just to fetch a single block

PR #218, opened 2023-10-31
don't require PK if service is not activated

siosw:
# Goals of PR

if a service like validator / sequencer is disabled, we should allow both the PK flag and the clef endpoint to be empty

### DIFF
--- a/services/sidecar/rollup/services/config.go
+++ b/services/sidecar/rollup/services/config.go
@@ -140,6 +140,9 @@ func (c SequencerConfig) GetTxMgrCfg() txmgr.Config               { return c.TxM
 
 // Validates the configuration.
 func (c SequencerConfig) validate() error {
+	if !c.IsEnabled {
+		return nil
+	}
 	if c.SecretKey == nil && c.ClefEndpoint == "" {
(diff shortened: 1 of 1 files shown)


services/sidecar/rollup/services/config.go
```diff
@@ -140,6 +140,9 @@ func (c SequencerConfig) GetTxMgrCfg() txmgr.Config               { return c.TxM
 
 // Validates the configuration.
 func (c SequencerConfig) validate() error {
+	if !c.IsEnabled {
+		return nil
+	}
 	if c.SecretKey == nil && c.ClefEndpoint == "" {
 		return fmt.Errorf("missing both secret key and clef endpoint (require at least one)")
```
macwis, line 147: If this allows to run without these, I feel like it would be better to convert this err return into a log.warning instead and return nil here, without an intro of validation on/off switch from config. But this is just a suggestion, maybe there is more to the whole approach to that problem that could be reconsidered.

macwis: Of if we disable validation than we should disable level up function before we call the `validate()` function.

macwis (approved):

PR #221, opened 2023-11-01
Improve scripts

ujvl:
# Goals of PR
- split `clean_*` scripts
- prompt user to confirm file overwrites
- moves `contracts/.env` out to `config/local_devnet`


config/example/.genesis.env
```diff
@@ -30,4 +30,5 @@ DEPLOYER_ADDRESS=0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC
 # What software stack to use to run the L1 network. Options: [geth|hardhat]
 L1_STACK=geth
 # L1 endpoint.
+# Note: if L1_STACK=hardhat, use host 127.0.0.1 (local) or 0.0.0.0 (docker)
```
MoonShiesty, line 33: should this comment say `172.17.0.1(docker)`?

we could also use the `L1_STACK` variable to set the default L1_ENDPOINT ip addr

MoonShiesty (approved):
looks good, left a single comment

config/example/.genesis.env
ujvl: - i think for HH specifically it wants 0.0.0.0
- yeah was gonna do that in a later PR

PR #230, opened 2023-11-06
Ensure L1Oracle block number increases monotonically

LEAFERx:
# Goals of PR

Core changes:

- Add a `require` state to ensure L1Oracle block number increases monotonically
  - So that the sequencer can not modify the oracle values in the middle of the batch undetected


macwis (approved):

ayushguptacse14 (approved):
Looks good to me

PR #231, opened 2023-11-06
Introduce new basic batch format

ujvl:
Introduces the following basic batch format (for version=`0x0`)
```
batch_data = RLP(sub_batch_0, sub_batch_1, ...)
# where...
sub_batch = [first_l2_block_num, tx_blocks]
# where...
# - first_l2_block_num: uint64
# - epoch_num: uint64
# - epoch_hash: uint256
# - tx_blocks: List[tx_block]
tx_block = [tx0, tx1, ...]
```
- Empty blocks are skipped
- Epoch changes are inferred via the oracle tx

Also:
- refactors the sidecar flags (many are obsolete/unused)
- `services/sidecar/rollup/services/rollup_config.go` is more-or-less taken from bedrock (with some fields removed)

Misc:
- Delete unused code
- Script fixes
- Remove extraneous sleeping in scripts (todo later: fix docker)
- Fixes some `fmt` imports


config/local_docker/.genesis.env
```diff
@@ -1,18 +1,13 @@
...
-
-ADDR=0.0.0.0
-L1_ENDPOINT=ws://$ADDR:8545
+BASE_ROLLUP_CFG_PATH=./base_sp_rollup.json
+ROLLUP_CFG_PATH=./sp_rollup.json
 
+L1_NETWORK=localhost
+L1_ENDPOINT=ws://0.0.0.0:8545
 L1_STACK=geth
-L1_DEPLOY=true
```
siosw, line 14: we should leave this for the docker config. or add a call to the deploy script in docker-compose if you prefer that

sbin/start_sp_magi.sh
```diff
@@ -53,8 +51,5 @@ CMD="$MAGI \
     $DEVNET_FLAGS \
     $SEQUENCER_FLAGS"
 
-# TODO: wait for config.json to be generated
-sleep 15
```
siosw, line 57: this will cause the docker compose to fail.
Currently trying to use like this [this](https://docs.docker.com/compose/compose-file/05-services/#depends_on) to start containers in the correct order without waiting in the scripts.

sbin/deploy_l1_contracts.sh
```diff
@@ -81,7 +81,7 @@ $SBIN/create_genesis.sh
 # Deploy contracts
 cd $CONTRACTS_DIR
 echo "Deploying l1 contracts..."
-npx hardhat deploy --network localhost
+npx hardhat deploy --network $NETWORK
```
siosw, line 84: this should be `L1_NETWORK` I think

config/local_docker/.genesis.env
```diff
@@ -1,18 +1,13 @@
 BASE_GENESIS_PATH=./base_genesis.json
 GENESIS_PATH=./genesis.json
-BASE_ROLLUP_CFG_PATH=./base_sp_local_devnet.json
-ROLLUP_CFG_PATH=./sp_local_devnet.json
-
-SEQUENCER_ADDRESS=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
-VALIDATOR_ADDRESS=0x70997970c51812dc3a010c7d01b50e0d17dc79c8
-DEPLOYER_ADDRESS=0x3C44CdDdB6a900fa2b585dd299e03d12FA4293BC
```
siosw, line 8: these are used by the L1 script to pre-fund accs

macwis:
When tried to run with docker-compose I have experienced a config args parsing error:
```
docker-sidecar-1  | Using sidecar dotenv: .sidecar.env
docker-sidecar-1  | starting sidecar with the following flags:
docker-sidecar-1  | --l1.endpoint ws://l1-geth:8545 --l2.endpoint http://sp-geth:4011 --protocol.rollup-cfg-path --protocol.rollup-addr 0xF6168876932289D073567f347121A267095f3DD6 --disseminator --disseminator.private-key --validator --validator.private-key
docker-sidecar-1  | failed to parse config: failed to parse protocol config: deploy config at --protocol.rollup-addr not found: open --protocol.rollup-addr: no such file or directory
```

Also I think there is a problem running other services, but that's probably coming from the env vars renaming with `_BIN`. Will investigate further.

config/local_docker/.genesis.env
ujvl: added a `-d` flag instead

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -43,99 +20,117 @@ type InvalidBlockError struct{ Msg string }
...
-	if err := w.WriteByte(TxBatchVersion()); err != nil {
-		return nil, err
+// TODO: this is not currently called anywhere but will be useful for testing.
+func DecodeBatch(data []byte) (interface{}, error) {
+	if len(data) == 0 {
+		return nil, &DecodeTxBatchError{"empty batch data"}
+	}
+	version := data[0]
+	switch version {
+	case 0:
```
LEAFERx, line 131: use the `v0` definition. also I would like to use all upper-case identifiers to mark it as a constant, like `V0` or `BATCH_V0`. separately, its better to use a table to dispatch the decode instead of `switch` statement, but since we only have 1 version and its for testing purpose to decode its fine now

services/sidecar/rollup/derivation/batch_v0_encoder.go
```diff
@@ -0,0 +1,169 @@
...
+	return &BatchV0Encoder{cfg, 0, &bytes.Buffer{}, newSubBatch()}
+}
+
+func (e *BatchV0Encoder) GetBatch(l1Head spTypes.BlockID) ([]byte, error) {
+	if e.timeout != 0 && l1Head.GetNumber() >= e.timeout {
+		data := e.currBuf.Bytes()
+		e.currBuf.Reset()
+		return data, nil
+	}
+	if e.currBuf.Len() < int(e.cfg.GetTargetBatchSize()) {
```
LEAFERx, line 44: if we only check the buffer length, is it possible that we submit batches that only contain few large transactions?

services/sidecar/rollup/derivation/batch_v0_encoder.go
```diff
@@ -0,0 +1,169 @@
...
+		return nil, errBatchTooSmall
+	}
```
LEAFERx, line 46:
```suggestion
	if e.currBuf.Len() < int(e.cfg.GetTargetBatchSize()) {
		if e.timeout == 0 || l1Head.GetNumber() < e.timeout {
			return nil, errBatchTooSmall
		}
	}
```

services/sidecar/rollup/derivation/batch_v0_encoder.go
```diff
@@ -0,0 +1,169 @@
...
+	if shouldCloseBatch {
+		return errBatchFull
+	}
+	// Skip intrinsically-derivable blocks.
+	if shouldSkipBlock {
+		log.Info("Skipping intrinsically-derivable block", "block#", block.NumberU64())
+		return nil
+	}
+	// Decode oracle tx.
+	l1Epoch, _, _, _, _, err := bridge.UnpackL1OracleInput(block.Transactions()[0])
```
LEAFERx, line 89: this is wrong: block not always contain oracle transaction.
also, should you create a new subbatch when l1Epoch changes so that the oracle transaction is always at the first block in the subbatch

services/sidecar/rollup/derivation/batch_v0_encoder.go
```diff
@@ -0,0 +1,169 @@
...
+	timeout := l1Epoch + e.cfg.GetSeqWindowSize() - e.cfg.GetSubSafetyMargin()
+	if e.timeout == 0 || e.timeout > timeout {
+		e.timeout = timeout
+	}
+}
+
+// Writes encoded sub-batch out to the buffer.
+func (e *BatchV0Encoder) closeSubBatch() error {
+	log.Info("Closing sub-batch...")
+	if err := rlp.Encode(e.currBuf, e.subBatch); err != nil {
```
LEAFERx, line 116: this is wrong: appending encoded `subBatch` to `currBuf` does not make it a list of `subBatch`. You should notice that `rlp(subBatch1) ++ rlp(subBatch2)`, `rlp([rlp(subBatch1), rlp(subBatch2)])` and `rlp([subBatch1, subBatch2])` are all different. The last format should be used. Therefore, you should maintain a `[]subBatch` field in the encoder. I'd suggest you create a test that test the encoding and decoding together to make sure it works.

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -43,99 +20,117 @@ type InvalidBlockError struct{ Msg string }
...
 }
 
-func (b *batchBuilder) LastAppended() types.BlockID { return b.lastAppended }
+type ProtocolConfig interface {
+	GetSeqWindowSize() uint64
+}
+
+type batchBuilder struct {
+	encoder       VersionedDataEncoder
+	pendingBlocks []ethTypes.Block
```
LEAFERx, line 35: any reason not using `[]*ethTypes.Block` here?

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -43,99 +20,117 @@ type InvalidBlockError struct{ Msg string }
...
-
-		// Enforce soft cap on batch size.
-		if rlp.ListSize(numBytes) > b.maxBatchSize {
-			log.Info("Reached max batch size", "numBytes", numBytes, "maxBatchSize", b.maxBatchSize, "numBlocks", len(blocks))
-			break
+	// Process pending blocks.
+	numProcessed := 0
+	for _, block := range b.pendingBlocks {
+		if err := b.encoder.ProcessBlock(&block); err != nil {
+			if errors.Is(err, errors.New("full batch")) {
```
LEAFERx, line 105: Use `errBatchFull`?

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -43,99 +20,117 @@ type InvalidBlockError struct{ Msg string }
...
-	batchAttrs, err := b.serializeToAttrs()
-	if err != nil {
-		return nil, fmt.Errorf("failed to serialize batch: %w", err)
+	// Process pending blocks.
+	if err := b.encodePending(); err != nil {
+		return nil, fmt.Errorf("failed to encode pending blocks into a new batch: %w", err)
 	}
-	batchData, err := encodeBatch(batchAttrs)
+	// Try to get a batch from encoder.
+	batch, err := b.encoder.GetBatch(l1Head)
```
LEAFERx, line 77: minor: move this line into the `encodePending()` function

LEAFERx (changes requested):
Didn't check the scripts. Several changes need to be made. But overall structure lgtm.

services/sidecar/rollup/services/disseminator/interface.go
```diff
@@ -19,18 +18,15 @@ type ForkChoiceResponse = engine.ForkChoiceResponse
 type BuildPayloadResponse = engine.ForkChoiceResponse
 
 type BatchBuilder interface {
-	Append(block derivation.DerivationBlock, header derivation.HeaderRef) error
-	LastAppended() types.BlockID
-	Build() ([]byte, error)
+	Enqueue(block *ethTypes.Block) error
```
macwis, line 21: interface declaration usually does not contain method argument names

config/local_docker/.genesis.env
ujvl: its in `.contracts.env`

services/sidecar/rollup/services/flags.go
```diff
@@ -26,154 +27,109 @@ func mergeFlagGroups(groups ...[]cli.Flag) []cli.Flag {
...
-		l2L1OracleBaseFeeSlotFlag,
-	}
-	sequencerCLIFlags = []cli.Flag{
-		sequencerEnableSequencerFlag,
-		sequencerAddrFlag,
-		sequencerSecretKeyFlag,
-		sequencerClefEndpointFlag,
-		sequencerPassphraseFlag,
-		sequencerSequencingIntervalFlag,
+	l1Flags       = []cli.Flag{l1EndpointFlag}
```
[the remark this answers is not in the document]
ujvl: will clean this up in another PR

services/sidecar/rollup/services/disseminator/interface.go
ujvl: can you cite a style guide for that? i thought its better to use so you can document the param usage

config/example/.genesis.env
```diff
@@ -25,10 +25,14 @@ SEQUENCER_ADDRESS=0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266
...
+# What network to deploy to (using hardhat). Optional.
+# Required only by `sbin/deploy_l1_contracts.sh`.
+L1_NETWORK=localhost
+# What software stack to use to run the L1 network. Optional.
+# Required only by `sbin/start_l1.sh`. Options: [geth|hardhat]
 L1_STACK=geth
 # L1 endpoint.
+# Required only by `sbin/start_l1.sh` and `sbin/create_genesis.sh`
 # Note: if L1_STACK=hardhat, use host 127.0.0.1 (local) or 0.0.0.0 (docker)
 L1_ENDPOINT=ws://172.17.0.1:8545
```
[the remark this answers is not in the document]
ujvl: maybe for geth in docker, not sure -- can fix later

services/sidecar/rollup/derivation/batch_v0_encoder.go
```diff
@@ -0,0 +1,162 @@
...
+	// If the timeout has been reached, the batch will be closed regardless of its current size.
+	if !force && e.currBuf.Len() < int(e.cfg.GetTargetBatchSize()) {
+		return nil, errBatchTooSmall
+	}
+	// Close the sub-batch if the batch can fit it.
+	if e.shouldCloseBatch() {
+		e.closeSubBatch()
+	}
+	// Encode all sub-batches (except the last).
+	data, err := rlp.EncodeToBytes(e.subBatches[len(e.subBatches)-1])
```
LEAFERx, line 43: missing `:`

services/sidecar/rollup/derivation/batch_v0_encoder.go
```diff
@@ -0,0 +1,162 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to encode batch: %w", err)
+	}
+	// Reset the buffer. Note: we don't reset the sub-batch in case it wasn't closed.
+	// In this case, it'll be used in the next batch.
+	e.currBuf.Reset()
+	return data, nil
```
LEAFERx, line 50: version is not encoded in `data`

LEAFERx (changes requested):

LEAFERx (approved):

services/sidecar/rollup/services/rollup_config.go
```diff
@@ -0,0 +1,255 @@
+package services
```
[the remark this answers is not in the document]
ujvl: todo: subsequent PR

PR #235, opened 2023-11-09
Version assertion

MoonShiesty:
# Goals of PR

Version assertion data, which currently is currently equivalent to vmHash.

- Describe changes here

Prepend a version byte to assertion data, and store as a blob and modify validator to publish new versioned assertion data blob format.

- Write notes here

These changes increase the size of `Rollup.sol` over the transaction size limit (by ~100 with optimizer on and 1 run). We'll have to split up `Rollup.sol` to deploy these changes.

The decoding versioned assertion data cannot revert or validators will be unable to challange assertions with invalid encoding.

### DIFF
--- a/contracts/src/IRollup.sol
+++ b/contracts/src/IRollup.sol
@@ -134,7 +134,7 @@ interface IRollup {
     }
 
     struct Assertion {
-        bytes32 stateHash; // Hash of execution state associated with assertion. Currently equiv to `vmHash`.
+        bytes32 stateCommitment; // Versioned execution state associated with assertion. Currently equiv to `keccak256(version || vmHash)`.
         uint256 blockNum; // Block number this assertion advanced to
         uint256 parent; // Parent assertion ID
         uint256 deadline; // Dispute deadline (L1 block number)
@@ -308,10 +308,10 @@ interface IRollup {
      *
      * Emits: `AssertionCreated` and `StakerStaked` events.
      *
-     * @param vmHash New VM hash.
-     * @param inboxSize Size of inbox corresponding to assertion (number of transactions).
+     * @param stateCommitment Currently keccak256(version || vmHash)
+     * @param blockNum Block number this assertion advances to.
      */
-    function createAssertion(bytes32 vmHash, uint256 inboxSize) external;
+    function createAssertion(bytes32 stateCommitment, uint256 blockNum) external;
 
     /**
      * @notice Initiates a dispute between a defender and challenger on an unconfirmed DA.
--- a/contracts/src/Rollup.sol
+++ b/contracts/src/Rollup.sol
@@ -59,7 +59,7 @@ abstract contract RollupBase is
 
     struct AssertionState {
         mapping(address => bool) stakers; // all stakers that have ever staked on this assertion.
-        mapping(bytes32 => bool) childStateHashes; // child assertion vm hashes
+        mapping(bytes32 => bool) childStateCommitments; // child state commitments
     }
 
     struct Zombie {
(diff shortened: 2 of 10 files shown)


contracts/src/Rollup.sol
```diff
@@ -57,9 +57,12 @@ abstract contract RollupBase is
     IDAProvider public daProvider;
     IVerifier public verifier;
 
+    // Current txBatch serialization version
+    uint8 public constant currentAssertionVersion = 0;
+
     struct AssertionState {
         mapping(address => bool) stakers; // all stakers that have ever staked on this assertion.
-        mapping(bytes32 => bool) childStateHashes; // child assertion vm hashes
+        mapping(bytes => bool) childAssertionData; // child assertion data
```
MoonShiesty, line 65: not sure if the key to this mapping should be the versioned blob, or leave it as a state hash

contracts/src/IRollup.sol
```diff
@@ -308,10 +314,10 @@ interface IRollup {
      *
      * Emits: `AssertionCreated` and `StakerStaked` events.
      *
-     * @param vmHash New VM hash.
-     * @param inboxSize Size of inbox corresponding to assertion (number of transactions).
+     * @param blockNum Block number this assertion advances to.
+     * @param assertionData Versioned assertion data
```
chalkandpaste, line 318: Can you leave a little bit of detail on what the structure of the data is. Maybe a ref somewhere or a "version int8 | vmHash int256" (assuming that's the structure). Maybe it's not terribly important as the data doesn't get used.

contracts/src/Rollup.sol
chalkandpaste: I think including the version is probably better, but the map may cost us more because of being arbitrary length byte array? @LEAFERx

MoonShiesty: this is reverted now, based on changes suggested by @ujvl

MoonShiesty:
@ujvl ready for review. I changed the implementation to version the `vmHash`, by encoding the assertion `vmHash`, as `StateRoot = keccak(version | vmHash)`, instead of using a blob data format.

I renamed the `vmHash` data to `StateRoot` because in the future version it might contain additional hashed data. Optimism/Magi calls it `OutputRoot`, not sure what the best name is.

services/sidecar/rollup/services/validator/stateRoot.go
ujvl, line 1: filename should be lowercase with underscore

services/sidecar/rollup/types/assertion.go
```diff
@@ -2,14 +2,14 @@ package types
...
-	"github.com/ethereum/go-ethereum/common"
 )
 
+type Bytes32 [32]byte
+
 // Assertion represents disputable assertion for L1 rollup contract
 type Assertion struct {
 	ID         *big.Int
-	VmHash     common.Hash
+	StateRoot  Bytes32
```
ujvl, line 12: name conflict with ethereum state root, we should rename this

MoonShiesty:
@ujvl ready for review - renamed to `state_commitment` and squashed into a single commit

services/sidecar/rollup/services/validator/validator.go
```diff
@@ -34,8 +35,8 @@ type Validator struct {
 }
 
 type assertionAttributes struct {
-	l2BlockNum uint64
-	l2VMHash   common.Hash
+	l2BlockNum      uint64
+	stateCommitment specularTypes.Bytes32
```
ujvl, line 39: nit for clarity
```suggestion
	l2StateCommitment specularTypes.Bytes32
```

ujvl (approved):

services/sidecar/rollup/services/validator/state_commitment.go
```diff
@@ -0,0 +1,70 @@
...
+func (o *StateCommitmentV0) Marshal() []byte {
+	var buf [64]byte
+	version := o.Version()
+	copy(buf[:32], version[:])
+	copy(buf[32:], o.l2VmHash[:])
+	return buf[:]
+}
+
+// StateCommitment returns the keccak256 hash of the marshaled L2 state commitment
+func StateCommitment(stateCommitment VersionedStateCommitment) types.Bytes32 {
```
ujvl, line 43: nit: consider `ToBytes32()` maybe

chalkandpaste (approved):

services/sidecar/rollup/services/validator/validator.go
MoonShiesty: done

PR #242, opened 2023-11-20
Fix disseminator input unpacking (and other misc bugfixes)

ujvl:
- Unpack inputs correctly
- Trigger sub-batch closing correctly (emptiness check was incorrect previously)
- Use the right l1 chain ID for geth
- update txmgr to the latest upstream (and add some tests)
- add some basic param validation to `config.go`

### DIFF
--- a/sbin/start_l1.sh
+++ b/sbin/start_l1.sh
@@ -86,6 +86,7 @@ echo "Starting L1..."
 if [ "$L1_STACK" = "geth" ]; then
     $L1_GETH_BIN \
       --dev \
+      --dev.period 4 \
       --http \
       --http.api eth,web3,net \
       --http.addr 0.0.0.0 \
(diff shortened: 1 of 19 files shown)


sbin/start_l1.sh
```diff
@@ -109,6 +112,10 @@ if [ "$L1_STACK" = "geth" ]; then
     $L1_GETH_BIN attach --exec \
       "eth.sendTransaction({ from: eth.coinbase, to: '"$DEPLOYER_ADDRESS"', value: web3.toWei(10000, 'ether') })" \
       $L1_ENDPOINT
+
+    # Wait for 1 block
+    echo "Waiting for chain progression..."
+    sleep 4
 elif [ "$L1_STACK" = "hardhat" ]; then
     echo "Using $CONTRACTS_DIR as HH proj"
     cd $CONTRACTS_DIR && npx hardhat node --no-deploy --hostname $L1_HOST --port $L1_PORT &> $LOG_FILE &
```
siosw, line 121: I think we should set the HH ID to 1337 here to avoid confusion

siosw (approved):

sbin/start_l1.sh
ujvl: will do in next one

PR #245, opened 2023-11-27
Fix sidecar, magi and scripts

LEAFERx:
# Goals of PR

Core changes:

- Fix sidecar disseminator batch encoding -- rlp won't serialize private fields.
- Import magi fixes. See https://github.com/SpecularL2/magi/pull/20.
- Fix scripts
  - Add validator address flag to cmd
  - Disable peer network in sp geth
  - Format `base_rollup.json` and remove `genesis.l2.hash` -- it will be generated by scripts
  - Use `genesis_hash.json` to fill in l2 genesis hash in `create_config.ts`.
    - #235 made the `vmHash` no longer the genesis hash.


config/local_devnet/base_sp_rollup.json
```diff
@@ -1,19 +1,18 @@
...
+        "l1": {
+            "hash": "0x0000000000000000000000000000000000000000000000000000000000000000",
+            "number": 0
+        },
+        "l2": {
+            "number": 0
+        },
+        "system_config": {
+            "batcherAddr": "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
+            "overhead": "0x0000000000000000000000000000000000000000000000000000000000000000",
```
siosw, line 12: I guess the idea is to set theses values here and then write them to `L1Oracle` in `create_l1_oracle_update_transaction` right?

siosw (approved):

contracts/scripts/config/create_config.ts
```diff
@@ -12,8 +12,9 @@ async function main() {
   const baseConfigPath = parseFlag("--in");
   const configPath = parseFlag("--out");
   const genesisPath = parseFlag("--genesis");
+  const genesisHashPath = parseFlag("--genesis-hash");
```
ujvl, line 15:
```suggestion
  const genesisHashPath = parseFlag("--genesis-hash-path");
```

ujvl (approved):

PR #246, opened 2023-11-27
read fee params from oracle contract

siosw:
# Goals of PR
- instead of reading the L1 fee scalar / overhead from config, read them from oracle storage slots
- the L1Oracle address is hardcoded in the library, matching Optimisms design

## Depends On
- SpecularL2/magi#22 to update the new oracle values
- SpecularL2/go-ethereum#8 to load hooks with the modified config


lib/el_golang_lib/hook/fee_hook.go
```diff
@@ -19,21 +19,26 @@ const (
 	txSignatureOverhead = 68
 )
 
+var (
+	BaseFeeSlot  = common.BigToHash(big.NewInt(253))
```
LEAFERx, line 23: can you use the bindings in `ops` to get the slot? the bindings has the storage layout now

LEAFERx (changes requested):
other changes lgtm

lib/el_golang_lib/hook/fee_hook.go
```diff
@@ -188,3 +193,27 @@ func zeroesAndOnes(data []byte) (int64, int64) {
...
+	overheadEntry, err := layout.GetStorageLayoutEntry("l1FeeOverhead")
+	if err != nil {
+		panic("could not get overhead storage slot")
+	}
+	scalarEntry, err := layout.GetStorageLayoutEntry("l1FeeScalar")
+	if err != nil {
+		panic("could not get scalar storage slot")
+	}
+
+	return common.BigToHash(new(big.Int).SetUint64(uint64(baseFeeEntry.Slot))),
```
LEAFERx, line 216: one thing im not sure is that why https://github.com/SpecularL2/specular/blob/dc1fac0b07720b8fdbe332a66369074503f86327/ops/bindings/solc/types.go#L81 (comes from optimism). should we change it to string and parse to common.Hex?

LEAFERx (approved):
lgtm lets wait for geth and magi

ujvl (approved):

PR #249, opened 2023-11-29
[sidecar] Flush batch encoder (bugfix)

ujvl:
- Batch encoder must be flushed when building the new batch
- Handle edge cases around the max batch size
- Drop batches that are too old so we don't waste time/gas trying to sequence them (they will be reorged out by the CL client)

Also added unit tests for encoder


services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -96,8 +96,15 @@ func (b *batchBuilder) Advance() {
...
 	force := b.timeout != 0 && l1Head.GetNumber() >= b.timeout
-	log.Info("Trying to get batch", "force?", force)
-	batch, err := b.encoder.GetBatch(force)
+	log.Info("Trying to get batch", "curr_l1#", l1Head.GetNumber(), "timeout_l1#", b.timeout, "force?", force)
+	batch, err := b.encoder.Flush(force)
+	if force {
+		// If it's too late to sequence, the batch should just be dropped entirely.
+		timeoutExceeded := l1Head.GetNumber() >= b.timeout+b.cfg.GetSubSafetyMargin()
+		if timeoutExceeded {
+			return nil, fmt.Errorf("batch timeout exceeded: %w", err)
```
LEAFERx, line 105: change `timeoutExceed` to batch too old? it is a bit confusing since there's a normal timeout

LEAFERx (approved):

services/sidecar/rollup/derivation/batch_builder.go
ujvl: hmm maybe will consider later; following `op-node` for now (took some of these checks from the channel mgr there)

PR #253, opened 2023-11-30
automate address configuration

MoonShiesty:
# Goals of PR

Automate the configuration of addresses by parsing the hardhat deployment files, and write the contract deployment addresses to a config file.

Core changes:

Writes deployment addresses to a `.deployments.env` file, removes and replaces references to hardcoded addresses.

Notes:

The address a contract is deployed to is `hash(deployer address | nonce)`, so we can't hardcode contract deployments.


macwis:
this is very useful, but why ts, why not make it part of `ops` tool?

MoonShiesty:
> this is very useful, but why ts, why not make it part of `ops` tool?

we hardhat to deploy our L1 contracts

the process is essentially: 

1. create l2 genesis (uses `ops`)
2. deploy l1 contracts (uses 'hardhat')
3. create rollup config (currently typescript `create_config.ts`, this PR currently also parses and writes the deployment addresses here)

i think if the plan is to move step 3 into the ops tool, then these changes should also go into `ops` tool, otherwise I think its simpler to leave these changes in the ts config script.

ujvl (approved):

PR #257, opened 2023-12-07
E2E deposit fixes

siosw:
# Goals of PR
- have a working deposit test
- add back deposit and transaction tests to github actions

## Changes
- append `deployments.env` to `.env` used in contracts dir
- remove `onlyProxy` guard in `L2Portal`
- add e2e specific config and fix addresses used in scripts
- change the deposit test to wait for the oracle to update & fund the L2Portal


sbin/run_e2e_tests.sh
```diff
@@ -99,6 +99,13 @@ sleep 1
 $SBIN/start_sidecar.sh &>proc3.out &
 sleep 1
 
+CONTRACTS_ENV=$CONTRACTS_DIR/.env
+until [ -f $CONTRACTS_ENV ]; do
+  echo "waiting for contracts/.env to be generated..."
+  sleep 4
+done
+cat .deployments.env >>$CONTRACTS_ENV
+
```
MoonShiesty, line 108: i think it makes more sense to put this in `deploy_l1_contracts.sh` since we copy `.contracts..env` and generate the `.deployments.env` there

MoonShiesty (approved):
looks good, awesome to see the deposit e2e test working!

MoonShiesty:
do we need all of these configuration workspaces? right now we have `example`, `local_devnet`, `local_docker` and `e2e`

chalkandpaste (approved):

PR #260, opened 2023-12-14
Fix withdraw

MoonShiesty:
this PR fixes the withdraw e2e test as well as additional cleanup

based on @siosw PR#259


sbin/start_l1.sh
```diff
@@ -103,7 +103,7 @@ echo "Starting L1..."
 if [ "$L1_STACK" = "geth" ]; then
   $L1_GETH_BIN \
     --dev \
-    --dev.period 0 \
+    --dev.period 2 \
```
MoonShiesty, line 106: withdraw test doesn't work with a dev period < 2

contracts/foundry.toml
```diff
@@ -12,8 +12,9 @@ cache_path  = 'forge-cache'
 allow_paths = [
     '../node_modules',
 ]
+via_ir = true
```
MoonShiesty, line 15: forge build never completes with this setting, so i cant run the tests

PR #262, opened 2023-12-15
historical state roots

siosw:
# Goals of PR

Core changes:
- keep track of historical state roots in oracle -> users have a bigger window of time to submit deposit finalization
- add prettier to github actions and pre-commit hooks
- remove unused retry test

The large diff is mostly from running the linter, the first commit contains all changes related to the historical state roots.


siosw:
closing since we want to handle this logic on the client side, IMO the bridge SDK should handle retries

PR #265, opened 2023-12-19
reorg-resistance and misc cleanup

ujvl:
# Goals of PR
- Provide validators re-org resistance when L1 chain re-orgs (matches optimism's checks). Mitigates a race condition: (1) validator fetches safe tip, (2) L1 reorgs batches out, (3) validator submits a now-incorrect assertion. This PR only provides the functionality, a subsequent PR will implement it.
- Require only a single validator staked on assertion for its confirmation (this is sufficient)
- Remove dead code in contracts (unused errors, functions)
- Remove dead code in sidecar


LEAFERx (commented):
lgtm

LEAFERx (approved):

PR #261, opened 2023-12-14
withdrawal and erc20 E2E test fixes

siosw:
# Goals of PR

based off of @MoonShiesty's #260 and adds fixes for 
- batches including first tx with empty to field
- allows compilation of smart contracts without IR

erc20 and withdraw are passing

PR #269, opened 2023-12-20
address comments for #261 

siosw:
# Goals of PR
- mostly adding docs and refactoring
- regarding [this comment](https://github.com/SpecularL2/specular/pull/261#discussion_r1428492313), i prefer an abstract contract here. makes it easier to share the structs across tests & implementation without the overhead of actually deploying a library instance on chain


siosw:
I feel like these changes should trigger the CI checks. thoughts?

contracts/src/libraries/Hashing.sol
```diff
@@ -30,8 +30,7 @@ library Hashing {
      */
     function createStateCommitmentV0(bytes32 l2BlockHash, bytes32 l2StateRoot) internal pure returns (bytes32) {
         // output v0 format is keccak256(version || l2BlockHash || l2StateRoot)
-        bytes memory stateCommitment = new bytes(32); // version 0 is a zero bytes32
-        stateCommitment = bytes.concat(stateCommitment, l2BlockHash, l2StateRoot);
+        bytes memory stateCommitment = bytes.concat(STATE_COMMITMENT_V0, l2BlockHash, l2StateRoot);
```
LEAFERx, line 33: Use `abi.encodePacked` instead

LEAFERx (changes requested):

contracts/src/libraries/Hashing.sol
siosw: can you explain why you prefer that? afaik `bytes.concat` was added to avoid using `abi.encodePacked here`
https://github.com/ethereum/solidity/issues/10903

services/sidecar/rollup/services/validator/state_commitment.go
```diff
@@ -66,7 +68,7 @@ func UnmarshalStateCommitment(data []byte) (VersionedStateCommitment, error) {
 }
 
 func unmarshalStateCommitmentV0(data []byte) (*StateCommitmentV0, error) {
-	if len(data) != 96 {
+	if len(data) != V0BufSize {
```
ujvl, line 74: also the constants below as well

ujvl (approved):

contracts/src/libraries/Hashing.sol
LEAFERx: oh nvm thought you're using the `Bytes` library.

LEAFERx (approved):

services/sidecar/rollup/services/validator/state_commitment.go
```diff
@@ -36,11 +42,11 @@ func (o *StateCommitmentV0) Version() Bytes32 {
...
 
 func (o *StateCommitmentV0) Marshal() []byte {
-	var buf [96]byte
+	var buf [V0BufSize]byte
 	version := o.Version()
-	copy(buf[:32], version[:])
-	copy(buf[32:64], o.l2BlockHash[:])
-	copy(buf[64:], o.l2StateRoot[:])
+	copy(buf[:V0VersionSize], version[:])
+	copy(buf[V0BlockHashOffset:V0BlockHashSize], o.l2BlockHash[:])
```
LEAFERx: `buf[V0BlockHashOffset:V0StateRootOffset]` instead

services/sidecar/rollup/services/validator/state_commitment.go
```diff
@@ -66,12 +72,12 @@ func UnmarshalStateCommitment(data []byte) (VersionedStateCommitment, error) {
...
 func unmarshalStateCommitmentV0(data []byte) (*StateCommitmentV0, error) {
-	if len(data) != 96 {
+	if len(data) != V0BufSize {
 		return nil, ErrInvalidStateCommitment
 	}
 	var l2State StateCommitmentV0
 	// data[:32] is the version
-	copy(l2State.l2BlockHash[:], data[32:64])
-	copy(l2State.l2StateRoot[:], data[64:])
+	copy(l2State.l2BlockHash[:], data[V0BlockHashOffset:V0BlockHashSize])
```
LEAFERx: `buf[V0BlockHashOffset:V0StateRootOffset]` instead

PR #272, opened 2023-12-24
[ops] Automate and dedup address config

ujvl:
- dedups address configuration (everything should be bootstrapped from single dotenv, currently `.contracts.env`)
- dedups base_rollup.json params (can be set automatically from genesis.json during config generation)
- removes dead/obsolete code in hardhat cfg and some other places
- cleans up scripts with common util
- adds wallet generation tool for convenience


config/e2e_test/base_sp_rollup.json
```diff
@@ -1,23 +1,12 @@
...
-        },
-        "l2": {
-            "number": 0
-        },
         "system_config": {
-            "batcherAddr": "0xf39fd6e51aad88f6f4ce6ab8827279cfffb92266",
             "overhead": "0x0000000000000000000000000000000000000000000000000000000000000000",
-            "scalar": "0x0000000000000000000000000000000000000000000000000000000000000000",
-            "gasLimit": 30000000
+            "scalar": "0x0000000000000000000000000000000000000000000000000000000000000000"
```
LEAFERx: maybe change the `overhead` and `scalar` to some non-zero value

contracts/hardhat.config.ts
```diff
@@ -117,29 +67,14 @@ const config: HardhatUserConfig = {
     clear: true,
   },
   namedAccounts: {
-    sequencer: {
```
LEAFERx, line 120: why not export sequencer as a named account. if we're using single source of truth we can only set the default to env.

ops/cmd/genesis/main.go
```diff
@@ -54,6 +56,10 @@ var Flags = []cli.Flag{
 		Name:  "l1-block",
 		Usage: "L1 block number",
 	},
+	&cli.StringFlag{
```
LEAFERx, line 60: alloc is not only about balance -- we might also want to set nonce/code/storage sometime later manually

ops/cmd/genesis/main.go
```diff
@@ -86,6 +92,15 @@ func GenerateSpecularGenesis(ctx *cli.Context) error {
 	if err != nil {
 		return err
 	}
+	if ctx.IsSet("extra-allocs") {
```
LEAFERx: where is this flag defined?

LEAFERx (commented):
Several questions:
- where is sequencer funded on l2 genesis? through `--allocs` flag?
- how can we fund a redeploy like `L2Portal`? will `--allocs` flag's balance value be overwritten by `BuildGenesis`? 
- I think maybe a better way is to still use the `allocs` json field instead of the flag, but allowing predefined replaces like writing genesis config as `{ "sequencerAddr" : { "balance": ... }}` and `sequencerAddr` will be replaced by the actual value from env.

i didn't check scripts and sidecar closely.

ops/cmd/genesis/main.go
ujvl: yeah this is kind of just a hack, we can clean this up later; but also if you want to you can still use `alloc` in the config—this is just for convenience to merge the dotenv vars in

ops/cmd/genesis/main.go
ujvl: supposed to be `alloc`, fixed

contracts/hardhat.config.ts
ujvl: iiuc it depends on the priv key then, which is unnecessary

config/e2e_test/base_sp_rollup.json
ujvl: do these values even matter, don't they get overloaded

also moved them to the config generation

chalkandpaste (approved):

PR #273, opened 2023-12-28
Fix test configs

ujvl:
# Goals of PR
- Add configuration back for tests, to be automated later.
- Add dependency in workflow
- Disseminator bug fix


MoonShiesty (approved):

.github/actions/foundry/action.yml
```diff
@@ -0,0 +1,21 @@
...
+        key: ${{ runner.os }}-foundry-${{ hashFiles('**/foundry.toml') }}
+        restore-keys: |
+          ${{ runner.os }}-foundry-
+
+    - name: Install Foundry
+      uses: foundry-rs/foundry-toolchain@v1
+      with:
+        version: nightly
+
+    - name: Run forge install
```
LEAFERx: forge install should be run under contracts dir. in e2e it is not the case

PR #274, opened 2023-12-30
Add historical roots to L1Oracle for deposit

LEAFERx:
The `L1Oracle` only stores the latest L1 root. Only deposit requests proved at that specific L1 block can be finalized. To improve UX, this pr adds historical state roots to `L1Oracle`, allowing users to prove at the nearest 256 L1 Oracle updates.

# Goals of PR

Core changes:

- Add `prevStateRoots` to `L1Oracle`
- Add `l1BlockNumber` param to `L2Portal::finalizeDepositTransaction`


contracts/src/bridge/L1Oracle.sol
```diff
@@ -98,6 +100,7 @@ contract L1Oracle is Initializable, UUPSUpgradeable, OwnableUpgradeable, Pausabl
         stateRoot = _stateRoot;
         l1FeeOverhead = _l1FeeOverhead;
         l1FeeScalar = _l1FeeScalar;
+        prevStateRoots[uint8(number % 256)] = _stateRoot;
```
ujvl: use an array?

also should we keep `stateRoot` since it's redundant

ujvl (approved):

contracts/src/bridge/L1Oracle.sol
siosw: I think mapping is more gas efficient.
But also think that we should only have a stateRoots mapping, not stateRoot and prevStateRoots

PR #278, opened 2024-01-02
Add bridge onboarding service

LEAFERx:
# Goals of PR

Core changes:

- Add `packages/onboarding` which implements an API endpoint that helps finalize first-time user's deposit request.
- Improve `sdk`.
- Update lockfile.


.changeset/pink-chefs-change.md
```diff
@@ -1,6 +1,7 @@
 ---
 "@specularl2/sdk": patch
 "@specularl2/contracts": patch
+"@specularl2/onbarding-services": patch
```
chalkandpaste: `onboarding-services`

LEAFERx: thanks, fixed

chalkandpaste (approved):
LGTM

siosw (approved):
looks good. don't think this is a big concern now but eventually I guess we'll need some sort of rate limit to avoid ppl draining the service

LEAFERx:
> looks good. don't think this is a big concern now but eventually I guess we'll need some sort of rate limit to avoid ppl draining the service

they will cost more to initiate deposit on L1 side than our services's cost of finalizing them. But yes rate limit should be introduced

LEAFERx:
A lot of `onboarding` code should be moved to `sdk` for general briding utilities.

PR #282, opened 2024-01-10
Parameters for Devnet/Testnet to fix "jerky"/"lurching" progression

chalkandpaste:
# Goals of PR

Core changes:

- Adds `MAX_SAFE_LAG` and `MAX_SAFE_LAG_DELTA` to the disseminator.
- Uses those values to trigger a batch to prevent the sequencer from being unable to generate new blocks. This check is in the `batch_builder` such that on a non-empty batch that has not been published will be forced to be published if the disseminator is lagging too far behind the sequencer's unsafe head.

Notes:

- There is an additional check that may be removed before this PR is accepted if we can figure out why L1Oracle stops getting updated. When it stops getting updated, we can enter a period with a lag with an truly empty batch. This prevents any further transaction from being processed on L2 (i.e., L2 cannot progress because unsafe head is too far from safe). A better fix may be in order.


config/example/.genesis.env
```diff
@@ -32,6 +32,8 @@ L1_NETWORK=localhost
 # Required for contract deployment. Options: [geth|hardhat]
 # Note: if L1_STACK="geth", the L1 chain ID must be set to 1337 (i.e. in the base rollup cfg)
 L1_STACK=geth
+# Sets the period of block creation
```
ujvl:
```suggestion
# Sets the L1 block speed (in terms of blocks-per-second)
```

services/sidecar/rollup/services/disseminator/batch_disseminator.go
```diff
@@ -87,7 +91,7 @@ func (d *BatchDisseminator) step(ctx context.Context) error {
 		}
 		return fmt.Errorf("failed to append to batch builder: %w", err)
 	}
-	if err := d.disseminateBatches(ctx); err != nil {
+	if err := d.disseminateBatches(ctx, end-start+1); err != nil {
```
ujvl: use safe

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -101,12 +102,15 @@ func (b *batchBuilder) Advance() {
 }
 
 // Tries to get the current batch.
-func (b *batchBuilder) getBatch(l1Head types.BlockID) ([]byte, error) {
+func (b *batchBuilder) getBatch(l1Head types.BlockID, currentLag uint64) ([]byte, error) {
+	// NOTE: 36 is # of l2 blocks within 6 l1 blocks/the confirmation period
+	lagTooGreat := currentLag + 36 >= b.cfg.GetMaxSafeLag()
```
ujvl: define param in config (or embed in the max safe lag directly for now)

also you can move this line to before L113

services/sidecar/rollup/derivation/batch_builder.go
```diff
@@ -101,12 +103,14 @@ func (b *batchBuilder) Advance() {
 }
 
 // Tries to get the current batch.
-func (b *batchBuilder) getBatch(l1Head types.BlockID) ([]byte, error) {
-	if b.encoder.IsEmpty() {
+func (b *batchBuilder) getBatch(l1Head types.BlockID, currentLag uint64) ([]byte, error) {
+	// NOTE: 36 is # of l2 blocks within 6 l1 blocks/the confirmation period
```
ujvl: nit: delete comment

ujvl (approved):

PR #290, opened 2024-01-16
Completed split of dockerfiles for dev vs local run

rhoop:
# Goals of PR

Core changes:

- Describe changes here

Notes:

- Write notes here

### DIFF
--- a/docker/docker-compose-latest-build.yml
+++ b/docker/docker-compose-latest-build.yml
@@ -16,6 +16,7 @@ services:
     tty: true
     healthcheck:
       test: bash -c "[ -f /tmp/.l1_started.sh.lock ]"
+      test: bash -c "[ -f /tmp/.l1_started.sh.lock ]"
       retries: 10
       interval: 5s
       start_period: 40s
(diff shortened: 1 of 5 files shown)


macwis (approved):

config/local_docker/.genesis.env
```diff
@@ -3,7 +3,6 @@ GENESIS_PATH=./genesis.json
 BASE_ROLLUP_CFG_PATH=./base_sp_rollup.json
 ROLLUP_CFG_PATH=./sp_rollup.json
 GENESIS_EXPORTED_HASH_PATH=./genesis_hash.json
-
 L1_NETWORK=localhost
```
chalkandpaste: We're missing a line here, defining `L1_PERIOD`, something I moved out of the scripts. Unless the scripts go reverted, this won't work. Insert/use `L1_PERIOD=12`.

docker/docker-compose-local-devel.yml
```diff
@@ -0,0 +1,107 @@
...
+services:
+
+  #------- Secrets
+  generator:
+    image: 792926601177.dkr.ecr.us-east-2.amazonaws.com/specular-platform:specular-latest
+    volumes:
+      - ../workspace:/specular/workspace
+      - ../sbin:/specular/sbin
+    working_dir: /specular/workspace
+    entrypoint: /bin/bash ../sbin/generate_secrets.sh -d -j -y
```
chalkandpaste: [Not something to address] One thing that we should keep in mind for the future here is we'll have to implement an upgrade cycle, and we'll probably, when CLI tool is implemented (@macwis) want to re-use the generated secrets.

chalkandpaste (approved):
Sans the 1 comment re: `L1_PERIOD`, LGTM

PR #293, opened 2024-01-25
use SPC to manage workspace directory

siosw:
# Goals of PR

when running locally, instead of manually copying a workspace dir, this should be managed through SPC

# TODO:
- [x] document how to download workspaces etc...
- [x] start all services through SPC
- [x] download SPC binary as part of the CI pipeline


macwis:
that's great, let's just get the spc binary with wget which should bring the e2e to work and we're set!

sbin/run_e2e_tests.sh
```diff
@@ -40,50 +40,53 @@ function ctrl_c() {
...
 done
-
-reqdotenv "deployments" ".deployments.env"
+reqdotenv "deployments" $DEPLOYMENTS_ENV
 
 # Start sp-geth
-$SBIN/start_sp_geth.sh -c &>proc.out &
-sleep 1
+$SBIN/start_sp_geth.sh -c &>$WORKSPACE_DIR/sp_geth.out &
+sleep 4
```
siosw, line 77: for some reasons the tests fail with `sleep 1` - i would like to get rid of the hardcoded waiting and rely on wait files. should I try implementing this for this PR or should this go in the backlog?

macwis: I think this is fine for now

macwis (approved):

macwis:
We might need to add the spc installation in the docker images in CI/CD pipelines, otherwise the command won't be available there.

PR #297, opened 2024-01-26
Docker Entrypoint Finalized

rhoop:
## Finalized scripts to work with K8
* Still need to add health checks via HTTP vs lock files to make this safe for unhealthy restarts and to remove EKS as a prerequisite 
* Entrypoint now injects all variables from the K8 mount points which will allow us to use different environments with a bit of DRY

### DIFF
--- a/sbin/entrypoint.sh
+++ b/sbin/entrypoint.sh
@@ -15,19 +15,25 @@ export INFURA_KEY=`cat infura_pk.txt`
 export DEPLOYER_PRIVATE_KEY=`cat deployer_pk.txt`
 export SEQUENCER_PRIVATE_KEY=`cat sequencer_pk.txt`
 export VALIDATOR_PRIV_KEY=`cat validator_pk.txt`
-
-printenv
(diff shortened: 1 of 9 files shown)


macwis (approved):

sbin/entrypoint.sh
```diff
@@ -1,7 +1,43 @@
...
+. .paths.env
+. .sidecar.env
+set +o allexport
+
+case "$1" in
+deploy)
+    # Run the main container command.
+    echo "Running deploy for genesis and JWT"
+    /specular/sbin/generate_jwt_secret.sh
+    /specular/sbin/deploy_l1_contracts.sh -y
```
macwis, line 32: we should detach completely the contracts deploy from the running sequence. we can do that later as long as this is not executed when run via infra

rhoop: We spoke about this today. I am going to separate the logic and add a done check to this.

chalkandpaste (approved):

PR #320, opened 2024-02-01
feat: SDK serviceBridge class implementation

sv-priv:
# Goals of PR

Core changes:

- Implemented all the core SDK functions in the cross-chain messenger class.

Notes:

- Test scripts are WIP


packages/sdk/index.ts
siosw, line 1: i think we can remove this

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -0,0 +1,572 @@
...
+        l2ChainId: NumberLike
+    }) {
+
+        this.l1SignerOrProvider = toSignerOrProvider(opts.l1SignerOrProvider)
+        this.l2SignerOrProvider = toSignerOrProvider(opts.l2SignerOrProvider)
+        this.l1RPCprovider = new ethers.providers.JsonRpcProvider(
+            "http://l1-geth:8545",
+        );
+        this.l2RPCprovider = new ethers.providers.JsonRpcProvider(
+            "http://sp-geth:401'",
```
siosw: port should be 4011

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -0,0 +1,572 @@
...
+    constructor(opts: {
+        l1SignerOrProvider: SignerOrProviderLike
+        l2SignerOrProvider: SignerOrProviderLike
+        l1ChainId: NumberLike
+        l2ChainId: NumberLike
+    }) {
+
+        this.l1SignerOrProvider = toSignerOrProvider(opts.l1SignerOrProvider)
+        this.l2SignerOrProvider = toSignerOrProvider(opts.l2SignerOrProvider)
+        this.l1RPCprovider = new ethers.providers.JsonRpcProvider(
```
siosw: RPC url will be different depending on the chain ID. so I think we should probably get it from constants but also allow users to provide their own

siosw: so basically `getUrlByNetworkId()`

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -0,0 +1,572 @@
...
+        }
+    }
+
+
+    /**
+    * Return the status of the message; 
+    * 
+    * @returns String
+    */
+    public async getMessageStatus(bridgeTxLogs: TransactionReceipt): Promise<String> {
```
siosw: i know you are already aware of this but imo this should return an enum, not a string.
and also indicate if the message has already been finalized (can be checked in the portal contract)

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -0,0 +1,572 @@
...
+         */
+        depositETH: async (
+            amount: NumberLike,
+            opts?: {
+                recipient?: AddressLike
+                l2GasLimit?: NumberLike
+            },
+        ): Promise<TransactionRequest> => {
+
+            const bridgeTx = await this.l1StandardBridge.populateTransaction.bridgeETH(200_000, [], {
```
siosw: i know `200_000` is taken from the e2e test, but it's a random amount.
maybe we can try a better default value? also imo users should be able to overwrite this

packages/sdk/src/cross-chain-messenger.ts
```diff
@@ -0,0 +1,572 @@
...
+
+            // get L1 deposit proof
+            const depositProof = await getDepositProof(
+                this.l1Portal.address,
+                depositEvent.args.depositHash,
+                hexlifyBlockNum(initialBlockNumber),
+                this.l1RPCprovider
+            );
+
+            waitUntilBlockConfirmed(this.l1Oracle, initialBlockNumber)
```
siosw: i think finalize deposit shouldn't wait for the block to be confirmed. it should just check if `l1Oracle.number >= initialBlockNumber` and throw an error if that's not true.
that way SDK users can implement whatever waiting logic they want

services/cl_clients/magi
siosw, line 1: this should probably be removed

packages/sdk/src/cross-chain-messenger.ts
sv-priv: so we should probably add another function to the SDK next to getMessageStatus() for waiting for the transaction? @siosw

packages/sdk/src/cross-chain-messenger.ts
sv-priv: This is assigned to a future task for enhancements to avoid sprint scope delays.

packages/sdk/scripts/test_deposit.ts
```diff
@@ -0,0 +1,59 @@
+import { ethers } from "ethers";
+import { formatEther } from "ethers/lib/utils";
+import { ServiceBridge } from '../src/service_bridge';
+import { delay } from "../src/utils";
+
+async function main() {
+
+    // TODO: put me in a .env file
```
chalkandpaste, line 8: Can we relocate this? V. useful a we ship the component.

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,686 @@
...
+    */
+    public async getDepositStatus(bridgeTxLogs: TransactionReceipt): Promise<MessageStatus> {
+
+        const desiredBlockNumber = bridgeTxLogs.blockNumber;
+        let currentOracleBlockNumber = await this.l1Oracle.number();
+
+        const depositEvent = this.l1Portal.interface.parseLog(bridgeTxLogs.logs[1]);
+        const depositHash = depositEvent.args.depositHash;
+
+        console.log("depositHash: ", depositHash)
```
chalkandpaste: Do these need to be here? Are they for debugging only or is there a benefit to the user?

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,686 @@
...
+
+        const depositFinalized = await this.l2Portal.finalizedDeposits(depositHash)
+
+        console.log("depositFinalized: ", depositFinalized)
```
chalkandpaste: --""--

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,686 @@
...
+    * @returns MessageStatus 
+    */
+    public async getWithdrawalStatus(bridgeTxLogs: TransactionReceipt): Promise<MessageStatus> {
+
+        const desiredBlockNumber = bridgeTxLogs.blockNumber;
+        let currentOracleBlockNumber = await this.l1Oracle.number();
+
+        const withdrawalEvent = this.l2Portal.interface.parseLog(bridgeTxLogs.logs[1]);
+        const withdrawalHash = withdrawalEvent.args.withdrawalHash;
+        console.log("withdrawalHash:", withdrawalHash)
```
chalkandpaste: --""--

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,686 @@
...
+        const withdrawalFinalized = await this.l1Portal.finalizedWithdrawals(withdrawalHash)
+        console.log("withdrawalFinalized", withdrawalFinalized)
```
chalkandpaste: --""--

packages/sdk/src/service_bridge.ts
sv-priv: Kinda both. But it's better to remove these because they are in the service_bridge class. Let's stick to the logs in the scripts only. Those are 100% to benefit the user.

packages/sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,97 @@
...
+    [ChainID in L2ChainID]: number
+} = {
+    [L2ChainID.SPECULAR]: 2 as const,
+    [L2ChainID.SPECULAR_HARDHAT_LOCAL]: 1 as const,
+}
+
+export const CHAIN_BLOCK_TIMES: {
+    [ChainID in L1ChainID]: number
+} = {
+    [L1ChainID.MAINNET]: 13 as const,
```
chalkandpaste: I'm not sure this is right. These should both be 12s, right?

packages/sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,97 @@
...
+    [L1ChainID.SEPOLIA]: 15 as const,
```
chalkandpaste: 12s?

packages/sdk/src/utils/constants.ts
sv-priv: We don't need them for now; just removed them.

chalkandpaste (approved):

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,681 @@
...
+
+        const depositEvent = this.l1Portal.interface.parseLog(bridgeTxLogs.logs[1]);
+        const depositHash = depositEvent.args.depositHash;
+
+
+        const depositFinalized = await this.l2Portal.finalizedDeposits(depositHash)
+
+        if (desiredBlockNumber <= currentOracleBlockNumber.toNumber()) {
+            return MessageStatus.READY;
+        } else if (depositFinalized) {
```
siosw: i think this needs to be checked first, `desiredBlockNumber <= currentOracleBlockNumber.toNumber()` is also true when message status is done

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,681 @@
...
+
+    /**
+    * Return the status of the withdrawal message; 
+    * 
+    * @returns MessageStatus 
+    */
+    public async getWithdrawalStatus(bridgeTxLogs: TransactionReceipt): Promise<MessageStatus> {
+
+        const desiredBlockNumber = bridgeTxLogs.blockNumber;
+        let currentOracleBlockNumber = await this.l1Oracle.number();
```
siosw: for the withdrawal status we need to check the block number of the first assertion, not of the l1 oracle 

see this https://github.com/SpecularL2/specular/blob/4411e58b514aebbcf6ab3d1193cbfd6ae6f0a834/contracts/scripts/e2e/utils.ts#L220

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,681 @@
...
+
+        const withdrawalEvent = this.l2Portal.interface.parseLog(bridgeTxLogs.logs[1]);
+        const withdrawalHash = withdrawalEvent.args.withdrawalHash;
+        const withdrawalFinalized = await this.l1Portal.finalizedWithdrawals(withdrawalHash)
+
+
+        if (desiredBlockNumber <= currentOracleBlockNumber.toNumber()) {
```
siosw: same as above, i think the order needs to be reversed here

siosw (commented):
just have minor comments on the message status methods, lgtm otherwise

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,681 @@
...
+            const depositProof = await getDepositProof(
+                this.l1Portal.address,
+                depositEvent.args.depositHash,
+                hexlifyBlockNum(initialBlockNumber),
+                this.l1RPCprovider
+            );
+            const blockno = await this.l1Oracle.number()
+            console.log("blockno", blockno.toNumber())
+
+            await waitUntilOracleBlock(this.l1Oracle, initialBlockNumber);
```
siosw: this needs to be removed still i think

packages/sdk/src/service_bridge.ts
siosw: you resolved this, but I still think this needs to be fixed

siosw (approved):

packages/sdk/.env.example
```diff
@@ -0,0 +1,2 @@
+L1_URL='http://localhost:8545'
```
LEAFERx: I guess you don't need `'` here?

packages/sdk/scripts/test_deposit.ts
```diff
@@ -0,0 +1,60 @@
...
+    const l1Url = 'http://localhost:8545'
```
LEAFERx: can you just use dotenv to load a testing env file like `.env.test`?

packages/sdk/scripts/test_deposit.ts
```diff
@@ -0,0 +1,60 @@
...
+    const l1Wallet = new ethers.Wallet("0x7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6", l1RpcProvider)
+    const l2Wallet = new ethers.Wallet('0x7c852118294e51e653712a81e05800f419141751be58f605c371e15141b007a6', l2RpcProvider)
+
+    const l1balance = await l1Wallet.getBalance()
+    console.log("L1 balancee: ", l1balance)
+
+    const serviceBridge = new ServiceBridge({
+        l1SignerOrProvider: l1Wallet, // l1 signer
+        l2SignerOrProvider: l2Wallet, // l2 signer
+        l1ChainId: 1337,
```
LEAFERx: these configs should also come from env file, let's fix them later

packages/sdk/src/interfaces/types.ts
```diff
@@ -0,0 +1,98 @@
...
+  // already finalized
+  DONE = 2
+}
+
+/**
+ * L2 network chain IDs.
+ */
+export enum L2ChainID {
+  SPECULAR = 93481,
+  SPECULAR_HARDHAT_LOCAL = 31337,
```
LEAFERx: local testing l2 chain id is `13527` right?

packages/sdk/src/service_bridge.ts
```diff
@@ -0,0 +1,661 @@
...
+    L1StandardBridge,
+    L2StandardBridge,
+    L2StandardBridge__factory,
+    Rollup,
+    Rollup__factory,
+} from "./types/contracts";
+import { l2PortalAddress } from './constants'
+import { JsonRpcProvider } from '@ethersproject/providers'
+
+//get the bindings
```
LEAFERx: wdym here?

packages/sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+    L2Contracts,
+    ContractsLike
+} from '../interfaces'
+
+// Docker
+// L1_RPC_URL = http://l1-geth:8545
+// L2_RPC_URL = http://sp-geth:4011
+
+const l1portalAddresses = {
+    sepolia: '0x457818a300BBa8E998889FAb2e8453BDebA238E0',
```
LEAFERx: these are not stable yet. use a configuration file like json or env file instead of hard-encoding them

packages/sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Loads the L1 contracts for a given network by the network Id.
+ *
+ * @param network The ID of the network to load the contracts for.
+ * @returns The L1 contracts for the given network.
+ */
+export const getL1ContractsByNetworkId = (network: number): L1Contracts => {
+    if (network == 11155111) {
```
LEAFERx: magic number

packages/sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+}
+
+/**
+ * Maps L1 chain IDs to the appropriate contract addresses of the given network.
+ */
+export const CONTRACT_ADDRESSES: {
+    [ChainID in L2ChainID]: ContractsLike
+} = {
+    [L2ChainID.SPECULAR]: {
+        l1: getL1ContractsByNetworkId(11155111),
```
LEAFERx: magic number

packages/sdk/src/utils/constants.ts
```diff
@@ -0,0 +1,80 @@
...
+        l2: DEFAULT_L2_CONTRACT_ADDRESSES,
+    },
+    [L2ChainID.SPECULAR_HARDHAT_LOCAL]: {
+        l1: getL1ContractsByNetworkId(31337),
```
LEAFERx: magic number

LEAFERx (approved):
can you run formatter before merge?

packages/sdk/scripts/test_deposit.ts
sv-priv: fixed them within this PR, low hanging fruit

packages/sdk/scripts/test_deposit.ts
sv-priv: sounds good

packages/sdk/src/interfaces/types.ts
sv-priv: yes, we don't used this enum apparently

packages/sdk/src/service_bridge.ts
sv-priv: i think this is outdated, will delete

packages/sdk/src/utils/constants.ts
sv-priv: leaving these like this the way they are, it's just the SDK only in the constants.ts file

PR #321, opened 2024-02-06
add configs for watcher delay

chalkandpaste:
# Goals of PR

Core changes:

- Implements configuration parameters for the new ChainWatcher delay parameter which allows for configurable rate limiting in checking for new L1 blocks

Notes:

- Be careful with the e2e parameter because we test with a block rate of 2s


rhoop (approved):
LGTM

config/example/.sp_magi.env
```diff
@@ -22,3 +22,4 @@ DEVNET=true
 SEQUENCER=true
 SEQUENCER_MAX_SAFE_LAG=200
 SEQUENCER_PK_FILE=./sequencer_pk.txt
+WATCHER_DELAY=4000
```
ujvl, line 25: nit: add a comment above this (this makes it look like it's a required param for sequencing)

ujvl (approved):

PR #325, opened 2024-02-13
bindings n genesis separation

macwis:
# Goals of PR

Core changes:

- Refactored bindings generation
- Refactored genesis generation and imports
- Refactored sidecar imports

Notes:

n/a


services/sidecar/go.mod
```diff
@@ -4,86 +4,74 @@ go 1.20
 
 replace github.com/ethereum/go-ethereum => ../el_clients/go-ethereum
 
+replace github.com/specularL2/specular/bindings-go => ../../bindings-go
+
 require (
 	github.com/avast/retry-go/v4 v4.3.3
-	github.com/ethereum/go-ethereum v1.12.2
+	github.com/ethereum/go-ethereum v1.10.26
```
LEAFERx: why change this version

lib/el_golang_lib/go.mod
```diff
@@ -4,11 +4,11 @@ go 1.20
 
 replace github.com/ethereum/go-ethereum => ../../services/el_clients/go-ethereum
 
-replace github.com/specularL2/specular/ops => ../../ops
+replace github.com/specularL2/specular/bindings-go => ../../bindings-go
 
 require (
-	github.com/ethereum/go-ethereum v1.13.2
-	github.com/specularL2/specular/ops v0.0.0-20231129092153-13f66846bd07
+	github.com/ethereum/go-ethereum v0.0.0-00010101000000-000000000000
```
LEAFERx, line 10: why change this version

bindings-go/go.mod
```diff
@@ -0,0 +1,38 @@
+module github.com/specularL2/specular/bindings-go
+
+go 1.20
+
+replace github.com/ethereum/go-ethereum => ../services/el_clients/go-ethereum
+
+require github.com/ethereum/go-ethereum v0.0.0-00010101000000-000000000000
```
LEAFERx, line 7: Set the version to `v1.13.2`?

.gitmodules
```diff
@@ -6,7 +6,7 @@
 	url = https://github.com/SpecularL2/blockscout
 [submodule "services/el_clients/go-ethereum"]
 	path = services/el_clients/go-ethereum
-	url = https://github.com/specularL2/go-ethereum
+	url = https://github.com/SpecularL2/go-ethereum.git
```
LEAFERx: why change this?

services/sidecar/go.mod
macwis: I think it just got autoupdated by mod tidy at some point or my IDE, but it makes no difference since we are replacing the import anyway

lib/el_golang_lib/go.mod
macwis: it's not me - go mod tidy does that when using replace

bindings-go/go.mod
macwis: go mod tidy does that when using replace

.gitmodules
macwis: already overwritten - I changed this to ssh path and IMHO we should do same with the rest of submodules, since (at least for me) HTTPS return:
```
remote: Support for password authentication was removed on August 13, 2021.
remote: Please see https://docs.github.com/en/get-started/getting-started-with-git/about-remote-repositories#cloning-with-https-urls for information on currently recommended modes of authentication.
fatal: Authentication failed for 'https://github.com/specularL2/go-ethereum/'
```

bindings-go/go.mod
macwis: changed back

services/sidecar/go.mod
macwis: changed back

PR #338, opened 2024-02-22
Fixed Imports for Docker and changed values for K8

rhoop:
Quick one line fixes already running in production


docker/build-master.Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN echo "deb [signed-by=/etc/apt/keyrings/nodesource.gpg] https://deb.nodesourc
 
 RUN apt-get update && apt-get install nodejs -y
 
-RUN apt install -y libudev-dev build-essential ca-certificates clang libpq-dev libssl-dev pkg-config lsof lld bash python3 make g++ musl-dev git
+RUN apt install -y libudev-dev build-essential ca-certificates clang libpq-dev libssl-dev pkg-config lsof lld bash python3 make g++ musl-dev git pgrep
```
macwis: @rhoop As we merged sth similar previously - the `pgrep` package does not exists and this will result in a failed build. There is another package which contains this `pcregrep` - I fixed this here: https://github.com/SpecularL2/specular/pull/333 , but @LEAFERx preferred to reverse the earlier PR merge instead of adding that.

macwis (changes requested):
Package install fix required.

LEAFERx (approved):

sv-priv (approved):

rhoop (commented):
This has been completed

ujvl (approved):

